PR #20, opened 2019-12-15
add Darwin as a platform so that calculation and assignment of a rawd…

sjswerdloff:
add Darwin as a platform so that calculation and assignment of a rawdvh will take place on MacOS
without the rawdvh, application crashes on MacOS
similar issue as Windows where the single process approach is required.

The error message on MacOS (Catalina/10.15.2) when using the multiprocess approach was:
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.

### DIFF
--- a/src/View/ProgressBar.py
+++ b/src/View/ProgressBar.py
@@ -65,6 +65,9 @@ class Extended(QtCore.QThread):
             elif self.platform == 'Windows':
                 self.raw_dvh = self.single_calc_dvhs(
                     self.dataset_rtss, self.dataset_rtdose, self.rois, self.my_callback)
+            elif self.platform == 'Darwin':
+                self.raw_dvh = self.single_calc_dvhs(
+                    self.dataset_rtss, self.dataset_rtdose, self.rois, self.my_callback)
             # self.raw_dvh = self.calc_dvhs(
             #     self.dataset_rtss, self.dataset_rtdose, self.rois, self.my_callback)
             self.dvh_x_y = self.converge_to_O_dvh(


src/View/ProgressBar.py
```diff
@@ -65,6 +65,9 @@ def run(self):
             elif self.platform == 'Windows':
```
SimonBiggs: Why not just:

```python
elif self.platform in ['Windows', 'Darwin']:
```

sjswerdloff: in [...] is a reasonable alternative.

I kept the same coding paradigm as the original.
If one is going to separate the two approaches and use "in" rather than == I believe it is easier on the reviewer and (next) coder to change *both* conditional expressions to use in [...].

Ideally, if one wants to improve the conceptual approach, rather than basing the conditional branches on platform the condition should be based on whatever disallows the use of the multi processing approach vs the single process approach, and then one has only a binary choice (rather than processing lists or cascading elif).

See https://docs.python.org/3/library/multiprocessing.html for the details, but...
_Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See bpo-33725._

 but...

_Warning The 'spawn' and 'forkserver' start methods cannot currently be used with “frozen” executables (i.e., binaries produced by packages like PyInstaller and cx_Freeze) on Unix. The 'fork' start method does work._

So... if you want to have this multiprocessing work in MacOS (and Windows, probably), one would need to have spawn be the method rather than fork (and use either the singleton approach in main, or the context approach locally).  But in doing that, one sacrifices being able to make it work in a frozen executable, which is a goal of the project if I understand correctly.

There doesn't seem to be an explanation of why fork() is unsafe on MacOS/Darwin and Windows.  So checking platform appears to be a good heuristic.

I suppose one could leave fork in use for Linux, switch to spawn for Windows (the warning is about freezing on Unix, of which Darwin is a variant, but it is not warning about Windows), but use single processing for Darwin.  Which... leaves you with an if/elif/elif situation. And if someone can figure out how to determine if the process is based on a frozen executable, then have that be part of the decision tree for Darwin.  Modern Macs have quad processors, so there is probably a decent kick in going with multiprocessing, given that the number of processes involved can easily be as many if not more than 4.

The code that does the multi-processing is
multiprocessing.Process(...) in calc_dvhs(...) and the underlying function being called by the multiple processes is multi_get_dvhs(...).  The calling function is doing one ROI per process.
I have seen RT SS with as few as three and as many as a few dozen ROIs.  As AI contouring advances, I would expect each organ to have it's own ROI, if not multiple ROIs distinguishing between subsets of organs.  Further, I would expect the number of processors/cores to just keep going up.  My work ("Physics") laptop has 16 cores.

Any solution that enables the application to work on MacOS is fine by me.
A solution that addresses the fundamental issue and optimizes the outcome would be superior to what I've offered so far.

SimonBiggs: Fair point about just working to the style already there.

Seeing the following:

https://github.com/sjswerdloff/OnkoDICOM/blob/add_DARWIN_platform_for_rawdvh/src/View/ProgressBar.py#L63-L64

https://github.com/sjswerdloff/OnkoDICOM/blob/add_DARWIN_platform_for_rawdvh/src/View/ProgressBar.py#L66-L67

https://github.com/sjswerdloff/OnkoDICOM/blob/add_DARWIN_platform_for_rawdvh/src/View/ProgressBar.py#L69-L70

In light of https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

... is very hard on the eyes ... ... makes me want to steer a little clear ...

PR #25, opened 2020-08-19
constructing PatientDictContainer and passing that in to MainWindow

sjswerdloff:
this is a bug fix (the app would core dump due to wrong signature for
creating MainWindow).
some reformatting based on my settings in (VS) code

### DIFF
--- a/src/Controller/interPageController.py
+++ b/src/Controller/interPageController.py
@@ -3,11 +3,14 @@
 #   This file handles all the processes done when opening a new patient or opening the program for the first time   #
 #                                                                                                                   #
 #####################################################################################################################
-from PyQt5.QtWidgets import QMessageBox, QDesktopWidget
 from shutil import which
+
+from PyQt5.QtWidgets import QDesktopWidget, QMessageBox
+
+from src.Model.PatientDictContainer import PatientDictContainer
+from src.View.Main_Page.ClinicalDataForm import *
 from src.View.Main_Page.mainPage import *
 from src.View.openpage import WelcomePage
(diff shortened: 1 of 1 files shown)


src/Controller/interPageController.py
```diff
@@ -299,15 +319,35 @@ def show_patient(self, path):
         """
         Display patient data
         """
+        patient_dict_container = PatientDictContainer(
```
sjswerdloff, line 322: This is the significant change (not just automated reformatting based on black and isort).  Create the PatientDictContainer

src/Controller/interPageController.py
```diff
@@ -299,15 +319,35 @@ def show_patient(self, path):
...
+            pixluts=self.bar_window.ext.dict_pixluts,
+        )
         # Loads the main window  by providing the necessary data obtained by the progress bar
-        self.patient_window = MainWindow(path, self.bar_window.ext.read_data_dict, self.bar_window.ext.file_names_dict,
-                                         self.bar_window.ext.rois, self.bar_window.ext.raw_dvh,
-                                         self.bar_window.ext.dvh_x_y, self.bar_window.ext.dict_raw_ContourData,
-                                         self.bar_window.ext.dict_NumPoints, self.bar_window.ext.dict_pixluts)
+        self.patient_window = MainWindow(
+            # path,
+            patient_dict_container,
```
sjswerdloff, line 336: and then pass in just the patient_dict_container.
It would be reasonable for a review comment to require that I eliminate all of the commented out arguments.
I left them there to make it clear what the change was, and why.

PR #29, opened 2020-08-25
use ndarray and np.matmul() instead of the deprecated matrix class

sjswerdloff:
reformatting by black

### DIFF
--- a/src/Model/ROI.py
+++ b/src/Model/ROI.py
@@ -1,12 +1,29 @@
 import collections
+import multiprocessing
 import sys
 import time
 
-from PyQt5.QtCore import QPoint
-from PyQt5.QtWidgets import QWidget, QApplication, QGraphicsScene, QGraphicsView, QLabel, QVBoxLayout, QMainWindow
-from PyQt5.QtGui import QPainter, QPainterPath, QPolygon, QPolygonF, QColor, QPixmap, QPen, QBrush
 import numpy as np
-import multiprocessing
+from PyQt5.QtCore import QPoint
+from PyQt5.QtGui import (
+    QBrush,
+    QColor,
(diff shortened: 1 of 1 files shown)


src/Model/ROI.py
```diff
@@ -110,20 +127,37 @@ def calculate_matrix(img_ds):
 
     # Equation C.7.6.2.1-1.
     # https://dicom.innolitics.com/ciods/rt-structure-set/roi-contour/30060039/30060040/30060050
-    matrix_M = np.matrix(
-        [[orientation[0] * dist_row, orientation[3] * dist_col, 0, position[0]],
-         [orientation[1] * dist_row, orientation[4] * dist_col, 0, position[1]],
-         [orientation[2] * dist_row, orientation[5] * dist_col, 0, position[2]],
-         [0, 0, 0, 1]]
+    matrix_M = np.ndarray(
```
sjswerdloff, line 130: crux of the change

src/Model/ROI.py
```diff
@@ -137,7 +171,7 @@ def get_pixluts(dict_ds):
     :return: a dictionary of transformation matrices
     """
     dict_pixluts = {}
-    non_img_type = ['rtdose', 'rtplan', 'rtss']
+    non_img_type = ["rtdose", "rtplan", "rtss"]
```
sjswerdloff, line 174: from here down is just reformatting by black

src/Model/ROI.py
```diff
@@ -1,12 +1,29 @@
 import collections
+import multiprocessing
```
sjswerdloff, line 2: imports did not change, just re-ordered by isort

Fincap (approved):

PR #33, opened 2020-09-02
Without RTSS, selecting ROI Delete option will crash the program

kylelw23:
When selecting ROI Delete option, if there are no RTSS, a guard message will display.
image


kylelw23:
This pull request pushes my commit from team2020 branch to master branch, what I was intended to do is push my commit from my branch to team2020.

PR #34, opened 2020-09-02
Team2020

Fincap:
DVH calculation can now be skipped when a RTSTRUCT and RTDOSE are identified.

As it stands, once in the main patient window there is no way to calculate DVHs. This option is to be implemented.


src/View/ProgressWindow.py
```diff
@@ -56,6 +58,17 @@ def update_progress(self, progress_update):
         self.text_field.setText(progress_update[0])
         self.progress_bar.setValue(progress_update[1])
 
+    def prompt_calc_dvh(self):
+        choice = QMessageBox.question(self, "Calculate DVHs?", "RTSTRUCT and RTDOSE datasets identified. Would you "
+                                                               "like you calculate DVHs? (This may take up to several "
```
sjswerdloff, line 63: change from:
Would you like you calculate DVHs
to:
Would you like to calculate DVHs

src/Model/PatientDictContainer.py
```diff
@@ -23,6 +23,12 @@ def __init__(self, path, dataset, filepaths, **kwargs):
         self.filepaths = filepaths
         self.additional_data = kwargs
 
+    def add(self, key, value):
```
sjswerdloff, line 26: implies silent overwrite.
If the PatientDictContainer is expected to be sort of immutable, the "additional_data" part is not going to act that way now that there is an add that overwrites.
Depending on the behaviour you want, it might be prudent to raise a KeyError if someone tries to add a key that is already there.

sjswerdloff (commented):
One minor message text change, but one significant design impact issue.
Please take a look at the design impact issue (PatientDictContainer.add(self,key,value))
and let's find a place to discuss that.  It might be fine the way it is, but it warrants discussion.

PR #35, opened 2020-09-02
Fix DICOM Tree tab displays files not part of dataset

kylelw23:
Due to the hardcoded nature of the DICOM tree tab, when files such as RTSTRUCT and RTDOSE are not loaded into the program they will still display as options in the dropdown menu. Selecting these will crash the program. Additionally, all image files will be labelled as CT slices, even if the slices are not CT images.

So I put combobox.addItem RTDose and RTStruct in mainwindow.has_rtss condition, that solved the above error.


had647:
All good for this commit. Thanks, I will merge it now

PR #38, opened 2020-09-03
Team2020

Fincap:
ROI renaming and deletion is now reflected in the DVH tab (including DVH export)

### DIFF
--- a/src/View/mainpage/StructureTab.py
+++ b/src/View/mainpage/StructureTab.py
@@ -147,9 +147,10 @@ class StructureTab(object):
 		changes is a tuple of (new_dataset, description_of_changes)
 		description_of_changes follows the format {"type_of_change": value_of_change}.
 		Examples: {"rename": ["TOOTH", "TEETH"]} represents that the TOOTH structure has been renamed to TEETH.
-		{"delete": "TEETH"} represents that the TEETH structure has been deleted.
+		{"delete": ["TEETH", "MAXILLA"]} represents that the TEETH and MAXILLA structures have been deleted.
 		"""
 
+		print(changes)
 		new_dataset = changes[0]
 
 		# If this is the first time the RTSS has been modified, create a modified indicator giving the user the option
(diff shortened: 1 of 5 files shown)


src/View/mainpage/StructureTab.py
```diff
@@ -141,12 +141,18 @@ def update_content(self):
 
 		self.scroll_area.setWidget(self.scroll_area_content)
 
-	def structure_modified(self, new_dataset):
+	def structure_modified(self, changes):
 		"""
 		Executes when a structure is renamed/deleted. Displays indicator that structure has changed.
+		changes is a tuple of (new_dataset, description_of_changes)
+		description_of_changes follows the format {"type_of_change": value_of_change}.
```
sjswerdloff, line 148: and are of type
``dict`` with key (type of change) having type ``str`` and value (value_of_change) having type ``list`` of ``str``
And the type_of_change drives how the list is to be interpreted.
While I figured it out, it took a bit.  The examples are great.  But it is nice to have a straight up specification of the data type involved, especially given that python is loosely typed and late binding.

src/View/mainpage/StructureWidget.py
```diff
@@ -8,7 +8,7 @@
 
 class StructureWidget(QtWidgets.QWidget):
 
-    structure_renamed = QtCore.pyqtSignal(Dataset)  # new PyDicom dataset
+    structure_renamed = QtCore.pyqtSignal(tuple)  # (new_dataset, change_description)
```
sjswerdloff, line 11: does it go without saying that the new_dataset, which is of type ``pydicom.dataset.Dataset`` is representing an RT Structure Set?  
At some point, people might want to use Segmentation objects (not something you need to worry about), but my point is that in addition to the python data type involved, in some of the code, it really does matter which DICOM object (SOP Class) is being represented by the pydicom.dataset.Dataset.  In other parts of the code, it really does not matter which SOP Class is involved.

PR #43, opened 2020-09-07
Add pseudonymisation and pymedphys requirement

sjswerdloff:
I've made the anonymisation capable of using pymedphys pseudonymisation.
However, I've also made it backward compatible in a python environment that doesn't have pymedphys in it, so the code falls back to using the previous style of anonymisation.
This was to avoid a breaking change on merge that would have caused *everyone* to have to update their environments right away.


src/Model/Anon.py
```diff
@@ -726,24 +668,76 @@ def anonymize(path, Datasets, FilePaths, rawdvh):
         first_dicom_object["PatientName"].repval
     )
 
-    if not file_previously_anonymised:
-        hashed_patient_id = _gen_md5_and_sha1_hash(Original_P_ID)
-        hashed_patient_name = _gen_md5_and_sha1_hash(patient_name_in_dataset)
+    # get the pname_ID ("Patient Name  + PatientID") before any anonymisation
```
sjswerdloff, line 671: note that this change is in the public anonymize() function, the diff viewer may make this appear to be in the private _anonymise_dicom_data()

Fincap (approved):

parise22 (approved):

PR #44, opened 2020-09-08
Removed a print statement used for testing

Fincap:
Removing print statement that was accidentally left in from previous commit.

### DIFF
--- a/src/View/mainpage/StructureTab.py
+++ b/src/View/mainpage/StructureTab.py
@@ -150,7 +150,6 @@ class StructureTab(object):
 		{"delete": ["TEETH", "MAXILLA"]} represents that the TEETH and MAXILLA structures have been deleted.
 		"""
 
-		print(changes)
 		new_dataset = changes[0]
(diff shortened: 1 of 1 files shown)


src/View/mainpage/StructureTab.py
```diff
@@ -150,7 +150,6 @@ def structure_modified(self, changes):
 		{"delete": ["TEETH", "MAXILLA"]} represents that the TEETH and MAXILLA structures have been deleted.
 		"""
 
-		print(changes)
```
sjswerdloff, line 153: consider use of
logging.debug() 
e.g.
logging.debug(changes)
if this would be useful for future debugging.
Some print statements are just about helping the programmer figure out "how it's supposed to work vs. how it is working right now" when the code is unfamiliar.
But if this is something that could be useful because "something got passed in that wasn't what was expected", then debug level logging can be useful (it's not about debugging the code "here", it's about debugging what came in).

PR #49, opened 2020-09-22
Adjustments to PatientBar alignment.

Fincap:
Minor alignment fixes to the PatientBar class.

Before:
Before

After:
After


parise22 (approved):
No problems, this drag and drop feature can be fully implemented once ROI Draw in complete.

PR #52, opened 2020-10-24
Main window restructure

Fincap:
The Main Window and all of its components have been completely rebuilt from the ground up. This is to ensure better adherence to OOP principles, and also to simplify the structure of the main window. Previously, components of the main window would directly modify the main window from their classes, which made it difficult to clearly understand how the whole thing pieced together. Now, all components are QWidgets and do not make any reference to their "parent" component (i.e. the main window). Additionally, the PatientDictContainer has been converted to a Singleton. This solves the memory leak issue. The components then reference this singleton to access the Model rather than the previous method of accessing instance attributes of the main window. This will make it far easier for new contributors to pick up and understand the structure more fully, which means that meaningful contributions can be made at a much quicker pace.

This PR also includes:
- Calculating DVHs from the main window if it was skipped when loading the patient.
- DICOM Tree now adapts to the dataset loaded in and doesn't present options which can crash the program.
- A fix to the segmentation fault caused upon closing the program.
- A fix to the memory leak upon opening a new patient.
- Relocation of the Draw and Delete ROI buttons to the Structures tab.
- Cleanup of toolbar and menubar actions, removing redundant and unused options.
- Added functionality to the previous unused Help action in the menubar: it now opens the OnkoDICOM website.

The restructure has resulted in the disappearance of two significant and notable features:
- The Clinical Data tab, and
- Metadata displayed on the DICOM View tab.

These features will be reimplemented in subsequent pull requests.


sjswerdloff:
While I will go through as much of this as I can, this large of a Pull Request is an example of "way too big" to review.
As a mental exercise, consider how this could have been done as a series of PRs, incrementally working towards the cumulative change.

legacy/src/Controller/interPageController.py
```diff
@@ -7,8 +7,7 @@
 
 from PyQt5.QtWidgets import QDesktopWidget
```
sjswerdloff, line 9: why are we maintaining legacy?

src/Controller/ActionHandler.py
```diff
@@ -0,0 +1,255 @@
+import sys
+
+from PyQt5 import QtGui, QtWidgets
+
+from src.Model.CalculateImages import get_pixmaps
+from src.Model.PatientDictContainer import PatientDictContainer
+
+
+class ActionHandler:
```
sjswerdloff, line 9: so what is this, why is it here, and how is it used?
While there are single comment lines below that say "X action" followed by code that is
self.zzz_X_action, there is no documentation that helps someone understand the bigger picture.

https://www.pythonforbeginners.com/basics/python-docstrings
or any equivalent introduction to docstrings (that's the first one that came up when I put
docstring python in the duckduckgo search box)

src/Controller/ActionHandler.py
```diff
@@ -0,0 +1,255 @@
...
+        if self.patient_dict_container.has_attribute("raw_dvh"):
+            self.main_page.dvh_tab.export_csv()
+        else:
+            QtWidgets.QMessageBox.information(self.main_page,
+                                              "Unable to export DVH",
+                                              "DVH cannot be exported as there is no DVH present.",
+                                              QtWidgets.QMessageBox.Ok)
+
+    def action_exit_handler(self):
+        sys.exit()
```
sjswerdloff, line 255: For a Qt based application, in general one should not call sys.exit().
See http://stackoverflow.com/questions/38283705/ddg#38285497

If the action_exit_handler is connected to the appropriate slot, then maybe this is OK, although one might expect some cleanup activity to be taking place, or at least a comment that says "this will get called on the Qt exit signal, so put cleanup calls in here".
I don't know if we have called in to C/C++ anywhere where some memory management or file I/O needs to get cleaned up before exiting, but that would be the classic scenario

src/Controller/AddOnOptionsController.py
```diff
@@ -362,7 +362,7 @@ def accepting(self):
             stream.close()
         # Close the Add-On Options Window after saving
 
-        if self.window.has_rtss:
+        if hasattr(self.window, 'structures_tab'):
```
sjswerdloff, line 365: is it an error if it doesn't have the structures_tab ?
if so, is it a big problem, or just something to be noted, perhaps in a log file.
See https://docs.python.org/3/howto/logging.html

src/Controller/GUIController.py
```diff
@@ -2,10 +2,11 @@
 from PyQt5 import QtCore, QtWidgets, QtGui
 from PyQt5.QtWidgets import QMessageBox
 
-from src.View.mainpage.MainPage import UIMainWindow
+from src.Model.PatientDictContainer import PatientDictContainer
```
sjswerdloff, line 5: we didn't really talk about sorting imports but, it's considered best practice.
See https://pypi.org/project/isort/ and https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement#20763446

src/Controller/GUIController.py
```diff
@@ -71,30 +71,35 @@ def open_new_patient(self):
...
                 if confirm_pyradi == QMessageBox.Yes:
-                    self.run_pyradiomics.emit(self.path, self.filepaths, self.hashed_path)
+                    self.run_pyradiomics.emit(path, filepaths, hashed_path)
                 if confirm_pyradi == QMessageBox.No:
                     pass
             else:
-                self.run_pyradiomics.emit(self.path, self.filepaths, self.hashed_path)
+                self.run_pyradiomics.emit(path, filepaths, hashed_path)
         else:
             exe_not_found = QMessageBox.information(self, "Error",
```
sjswerdloff, line 91: And this is exactly where you want to put in a mention of ensuring that the PATH includes the directory where the executable is installed.

src/Controller/ROIOptionsController.py
```diff
@@ -3,39 +3,47 @@
...
-    def __init__(self, window, rois, dataset_rtss):
+
+    def __init__(self, rois, dataset_rtss):
         super(RoiDeleteOptions, self).__init__()
 
-        self.window = window
         self.setup_ui(self, rois, dataset_rtss, self.deleting_rois_structure_tuple)
+
+
 # The class that will be called by the main page to access the ROI Options controller
```
sjswerdloff, line 17: This is a useful comment, but it should be done as a docstring.
vscode has plugins to auto-populate docstrings, perhaps PyCharm has something similar.

src/Model/ImageLoading.py
```diff
@@ -220,7 +220,7 @@ def get_roi_info(dataset_rtss):
     return dict_roi
 
 
-def calc_dvhs(dataset_rtss, dataset_rtdose, rois, interrupt_flag, dose_limit=None):
+def calc_dvhs(dataset_rtss, dataset_rtdose, rois, interrupt_flag, dose_limit=None, progress_callback=None):
```
sjswerdloff, line 223: failed to update the docstring to add this optional parameter.
Knowing the type of the progress callback would be useful

src/Model/Isodose.py
```diff
@@ -95,4 +95,26 @@ def get_dose_grid(rtd, z=0):
...
\ No newline at end of file
+        return np.array([])
+
+
+def calculate_rxdose(rtplan):
+    rxdose = 1
+    if ('DoseReferenceSequence' in rtplan and
+            rtplan.DoseReferenceSequence[0].DoseReferenceStructureType and
+            rtplan.DoseReferenceSequence[0].TargetPrescriptionDose):
+        rxdose = rtplan.DoseReferenceSequence[0].TargetPrescriptionDose * 100
```
sjswerdloff, line 106: Magic Number.
CGRAY_PER_GRAY = 100 
or 
GRAY_TO_CGRAY_SCALE_FACTOR = 100
which also suggests that if someone writes this out to a DICOM object somewhere, they had better get the scaling right or you've just killed someone.
I would refactor by renaming the variable from "rxdose" to "rx_dose_in_cgray".

sjswerdloff: The reason why it doesn't have to be labelled tht way in the pydicom elements is that the definition of which units are used is in the Standard.

src/Model/PatientDictContainer.py
```diff
@@ -7,25 +12,60 @@
...
+from src.Model.Singleton import Singleton
+
 
+class PatientDictContainer(metaclass=Singleton):
 
-class PatientDictContainer:
+    def __init__(self):
+        # Initialize base requirements
+        self.path = None                # The path of the loaded directory.
+        self.dataset = None             # Dictionary of PyDicom dataset objects.
```
sjswerdloff, line 23: suggest using type hints.
And the name is problematic.
pydicom.dataset.Dataset is the underlying type, and it's referred to as a dataset.
Here, dataset means what the comment says (so at least it's commented, but it wouldn't hurt to use type hints, e.g

from typing import Dict
...
self.dataset : Dict = None
or alternatively to use comment based type hinting

src/Model/PatientDictContainer.py
```diff
@@ -7,25 +12,60 @@
...
+        self.additional_data = None     # Any additional values that are required (e.g. rois, raw_dvh, raw_contour, etc)
+
+    def set_base_values(self, path, dataset, filepaths, **kwargs):
         """
-        :param path: The path of the loaded directory.
-        :param dataset: Dictionary where keys are slice number/RT modality and values are PyDicom dataset objects.
-        :param filepaths: Dictionary where keys are slice number/RT modality and values are filepaths.
-        :param kwargs: Any additional values that are required (e.g. rois, raw_dvh, raw_contour, etc)
+        Used to initialize the data on the creation of a new patient.
+        :param path:
```
sjswerdloff, line 31: the descriptions were discarded.  That's problematic.

src/Model/PatientDictContainer.py
```diff
@@ -7,25 +12,60 @@
...
+    def __init__(self):
+        # Initialize base requirements
+        self.path = None                # The path of the loaded directory.
+        self.dataset = None             # Dictionary of PyDicom dataset objects.
+        self.filepaths = None           # Dictionary of filepaths.
 
-    def __init__(self, path, dataset, filepaths, **kwargs):
+        self.additional_data = None     # Any additional values that are required (e.g. rois, raw_dvh, raw_contour, etc)
+
+    def set_base_values(self, path, dataset, filepaths, **kwargs):
```
sjswerdloff, line 28: why "base"?
Are these the initial values?
Is this what gets called when loading a patient's data at the beginning?

src/Model/PatientDictContainer.py
```diff
@@ -7,25 +12,60 @@
...
+        self.path = None
+        self.dataset = None
+        self.filepaths = None
+        self.additional_data = None
+
+    def is_empty(self):
+        """
+        :return: True if class is empty
+        """
+        if self.path is not None:
```
sjswerdloff, line 55: it's a nit, but could this have been

if self.path or self.dataset or self.filepaths or self.additional_data is not None:
  return False

less code, easier to read.

src/Model/PatientDictContainer.py
```diff
@@ -7,25 +12,60 @@
...
+        if self.dataset is not None:
+            return False
+        if self.filepaths is not None:
+            return False
+        if self.additional_data is not None:
+            return False
+
+        return True
+
+    def set(self, key, value):
```
sjswerdloff, line 66: does this have to be public?
Is it something you are deliberately exposing so any new key / value pair can be provided by any consumer of the class?

src/Model/ROI.py
```diff
@@ -318,6 +318,13 @@ def get_roi_contour_pixel(dict_raw_ContourData, roi_list, dict_pixluts):
...
 
 
+def ordered_list_rois(rois):
+    res = []
+    for id, value in rois.items():
+        res.append(id)
+    return sorted(res)
+
+
 # class Test(QWidget):
```
sjswerdloff, line 328: Is there a reason to keep this commented out Test class?

src/View/ImageLoader.py
```diff
@@ -39,14 +39,19 @@ def load(self, interrupt_flag, progress_callback):
         # if not ImageLoading.is_dataset_dicom_rt(read_data_dict):
         #    raise ImageLoading.NotRTSetError
 
+        # Initialize the PatientDictContainer singleton.
```
sjswerdloff, line 42: This isn't initializing the PatientDictContainer.
It's getting the singleton.  The singleton is responsible for initialising itself.
After this gets the singleton it is (re)populating the singleton, or perhaps re-setting it's contents.

src/View/mainpage/DrawROIWindow.py
```diff
@@ -1,22 +1,21 @@
...
+from PyQt5.QtWidgets import QMessageBox, QHBoxLayout, QLineEdit, QSizePolicy, QPushButton, QDialog, QListWidget, QGraphicsPixmapItem, QGraphicsEllipseItem
 from PyQt5.Qt import Qt
-import os
-from src.Model import ROI
-from src.View.mainpage.DicomView import *
-import matplotlib.pyplot as plt
 
+from src.Controller.MainPageController import MainPageCallClass
+from src.Model.PatientDictContainer import PatientDictContainer
+from legacy.src.View.mainpage.DicomView import *
```
sjswerdloff, line 10: are we relying on legacy?
If not, get rid of this import, or else it makes it seem like we have to keep legacy, and odds are we don't want to.

src/View/mainpage/DrawROIWindow.py
```diff
@@ -548,10 +541,7 @@ def update_metadata(self):
             slice_pos = imagePosPatientCoordinates[2]
 
         # For formatting
-        if self.window.zoom == 1:
-            zoom = 1
-        else:
-            zoom = float("{0:.2f}".format(self.window.zoom))
+        zoom = 1
```
sjswerdloff, line 544: deliberate elimination of support for window zoom?
That's OK, I'd have to look at the commit comments to know why.

src/View/mainpage/IsodoseTab.py
```diff
@@ -1,189 +1,160 @@
...
-		color_square_label = QtWidgets.QLabel()
-		color_square_pix = QtGui.QPixmap(15, 15)
-		color_square_pix.fill(QtGui.QColor(red, green, blue))
-		color_square_label.setPixmap(color_square_pix)
-		return color_square_label
\ No newline at end of file
+from PyQt5 import QtWidgets, QtGui, QtCore
+
+from src.Model.PatientDictContainer import PatientDictContainer
+
```
sjswerdloff, line 4: the docstring got stripped out.
They should be put back in and updated if need be.

src/View/mainpage/IsodoseTab.py
```diff
@@ -1,189 +1,160 @@
...
+    def init_color_squares(self):
+        list_of_squares = []
+        for key, value in self.color_dict.items():
+            list_of_squares.append(self.draw_color_square(value))
+
+        return list_of_squares
+
+    def init_checkboxes(self):
+        list_of_checkboxes = []
+        # Values of Isodoses
```
sjswerdloff, line 63: this reeks of "should be refactored" because of the DRY ("Don't Repeat Yourself") principle.

src/View/mainpage/MainPage.py
```diff
@@ -1,349 +1,215 @@
...
-from src.View.mainpage.DicomView import *
-from src.View.mainpage.DVH import *
-from src.View.mainpage.DicomTree import *
-from src.View.mainpage.MenuBar import *
-from src.View.mainpage.resources_rc import *
-
-class UIMainWindow(object):
-
-    # To initiate progress bar for pyradiomics through anonymization
+from src.Model.CalculateImages import convert_raw_data, get_pixmaps
```
sjswerdloff, line 9: excellent.  got rid of import *!

src/View/mainpage/MenuBar.py
```diff
@@ -1,528 +1,46 @@
-import sys
-from copy import deepcopy
+import webbrowser
 
-from PyQt5 import QtCore, QtGui, QtWidgets
-from PyQt5.QtWidgets import QMessageBox
+from PyQt5 import QtWidgets, QtCore, QtGui
 
-from src.Model.CalculateImages import *
+from src.Controller.ActionHandler import ActionHandler
```
sjswerdloff, line 5: presumably most of what had been in this module got factored out to ActionHandler?

src/View/mainpage/NewStructureTab.py
```diff
@@ -0,0 +1,279 @@
...
+from pandas import np
+
+from src.Controller.ROIOptionsController import RoiDeleteOptions, ROIDelOption, ROIDrawOption
+from src.Model import ImageLoading
+from src.Model.PatientDictContainer import PatientDictContainer
+from src.Model.ROI import ordered_list_rois
+from src.View.mainpage.StructureWidget import StructureWidget
+
+
+class NewStructureTab(QtWidgets.QWidget):
```
sjswerdloff, line 15: let's hope there isn't a NewNewStructureTab in the future.
Maybe the name could contain something that indicates how this is different from the "old" StructureTab ?

src/View/mainpage/NewStructureTab.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def roi_delete_clicked(self):
+        self.roi_delete_handler.show_roi_delete_options()
+
+    def roi_draw_clicked(self):
+        self.roi_draw_handler.show_roi_draw_options()
+
+    def structure_modified(self, changes):
+        """
+        Executes when a structure is renamed/deleted. Displays indicator that structure has changed.
```
sjswerdloff, line 177: This is a good docstring

src/View/mainpage/PatientBar.py
```diff
@@ -1,169 +1,137 @@
...
+
+        # Set layout
+        self.patient_bar_layout.addWidget(self.icon)
+        self.patient_bar_layout.addWidget(self.name_widget)
+        self.patient_bar_layout.addWidget(self.id_widget)
+        self.patient_bar_layout.addWidget(self.gender_widget)
+        self.patient_bar_layout.addWidget(self.dob_widget)
+        self.setLayout(self.patient_bar_layout)
+
+    def create_patient_name(self):
```
sjswerdloff, line 50: is this creating a patient name?
or creating a widget that presents the patient name?

src/View/mainpage/PatientBar.py
```diff
@@ -1,169 +1,137 @@
...
+        name_layout.addWidget(name_label)
+
+        # Value
+        name_value = QtWidgets.QLabel(self.name_widget)
+        name_value.setAlignment(QtCore.Qt.AlignLeft)
+        name_value.setFont(QtGui.QFont("Laksaman", pointSize=10))
+        name_value.setText(self.basic_info['name'])
+        name_layout.addWidget(name_value)
+
+    def create_patient_id(self):
```
sjswerdloff, line 72: it looks like there might be some room to refactor out some of the common code.

sjswerdloff (approved):
There's nothing here that I would know to say "this must be fixed".
However, I suggest creating a set of issues in response to my comments so that those can be followed up over time.  My preference is for the issues to be in GitHub.

src/Controller/AddOnOptionsController.py
Fincap: Currently, the structures_tab is only created when an RTSTRUCT is present. It's not an error if there's no structures tab. If an RTSTRUCT is created during runtime, then a structures tab will also be created.

src/Model/PatientDictContainer.py
Fincap: It is deliberately public so that any class interacting with the object can make changes to the model.

PR #53, opened 2020-10-24
Code cleanup

Fincap:
This addresses some of the issues brought up in the comments of #52. Note that there is no additional documentation present in this pull request, that will come in a later merge.


src/Controller/AddOnOptionsController.py
```diff
@@ -4,16 +4,15 @@
...
 ######################################################################################################
 import csv
-import sys
 import webbrowser
 from collections import deque
 
 from PyQt5.QtWidgets import QFileDialog
 
-from src.data.csv import *
 from src.View.AddOnOptions import *
```
sjswerdloff, line 12: best to eliminate any import * and make it explicit so that we know what comes from where, and to avoid function or class name clashes

src/View/mainpage/DicomView.py
```diff
@@ -155,7 +155,7 @@ def isodose_display(self):
             # sort selected_doses in ascending order so that the high dose isodose washes
             # paint over the lower dose isodose washes
             for sd in sorted(self.patient_dict_container.get("selected_doses")):
-                dose_level = sd * self.patient_dict_container.get("rxdose") / \
+                dose_level = sd * self.patient_dict_container.get("rx_dose_in_cgray") / \
                              (dataset_rtdose.DoseGridScaling * 10000)
```
sjswerdloff, line 159: 10000 is a magic number.
What does it represent?  What are it's units (or is it unitless) ?

src/View/mainpage/PatientBar.py
```diff
@@ -21,23 +47,19 @@ def __init__(self):
         self.icon.setMaximumWidth(30)
         self.icon.setGeometry(QtCore.QRect(10, 0, 0, 0))
         self.icon.setText("")
-        self.icon.setPixmap(QtGui.QPixmap(":/images/Icon/patient.png"))
+        self.icon.setPixmap(QtGui.QPixmap("src/Icon/patient.png"))
 
         # Create patient name
-        self.name_widget = QtWidgets.QWidget()
-        self.create_patient_name()
+        self.name_widget = create_info_widget("Name", self.basic_info['name'])
```
sjswerdloff, line 53: You seem to have lost the Geometry information.
Is that OK?
I think you want to pass in the Geometry, at least optionally (named parameter).

sjswerdloff (changes requested):
I am concerned that you have lost necessary Geometry information while refactoring the information widgets.
And I strongly urge replacing the magic number used for dose level with a constant with a useful name (not just "SCALING_CONSTANT", but something that explains why that particular value.  At a minimum have a comment above the CONSTANT that describes it.

Some of this had to be all at once (altered signatures).
But alot of it could have been done separately.
Given the individual changes were relatively small, I would have preferred to have the import cleanup (separate from the module renaming) done as one PR, and then the variable/signature/module renaming as a separate PR.
At least that way each PR has a single purpose (beyond "addressing comments")

If the loss of Geometry information in the information widgets (don't remember which file, sorry, too many files involved!) is not an issue, just say so in a response and I'll Approve.

PR #58, opened 2020-10-26
Isodose tab reworked to reduce code repetition

Fincap:
No functional differences.

### DIFF
--- a/src/View/mainpage/IsodoseTab.py
+++ b/src/View/mainpage/IsodoseTab.py
@@ -81,6 +81,7 @@ class IsodoseTab(QtWidgets.QWidget):
             if first_iteration:
                 checkbox = QtWidgets.QCheckBox("%s %% / %s cGy [Max]" % (str(isodose_percentages[i]),
                                                                          str(list_of_doses[i])))
+                first_iteration = False
             else:
                 checkbox = QtWidgets.QCheckBox("%s %% / %s cGy" % (str(isodose_percentages[i]), str(list_of_doses[i])))
             checkbox.clicked.connect(lambda state, text=isodose_percentages[i]: self.checked_dose(state, text))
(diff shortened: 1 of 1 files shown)


src/View/mainpage/IsodoseTab.py
```diff
@@ -33,21 +35,22 @@ def init_layout(self):
 
     def init_color_isod(self):
         """
-        Create a dictionary containing the colors for each isodose.
+        Create a list containing the colors for each isodose.
 
-        :return: Dictionary where the key is the percentage of isodose and the value a QColor object.
+        :return: List of QColor objects that correspond to each isodose as defined outside the class..
```
sjswerdloff, line 40: This is still a Dictionary

src/View/mainpage/IsodoseTab.py
```diff
@@ -67,65 +70,26 @@ def init_checkboxes(self):
...
-        list_of_checkboxes.append(checkbox3)
-        list_of_checkboxes.append(checkbox4)
-        list_of_checkboxes.append(checkbox5)
-        list_of_checkboxes.append(checkbox6)
-        list_of_checkboxes.append(checkbox7)
-        list_of_checkboxes.append(checkbox8)
-        list_of_checkboxes.append(checkbox9)
-        list_of_checkboxes.append(checkbox10)
+        first_iteration = True
+        for i in range(10):
```
sjswerdloff, line 80: or have one line for i = 0 and then range(1:10) for the rest.

while the following is not ideal for this situation, I've learned about enumerate
```
for i, isodose_percent in enumerate(isodose_percentages):
  print(i, isodose_percent) # or whatever else you need the count/index and value for...
```

sjswerdloff (approved):
but please do take a look at the comments

src/View/mainpage/IsodoseTab.py
Fincap: Whoops, I changed the type to a list while testing, and then changed it back to a dictionary but forgot to update the documentation. Will fix.

src/View/mainpage/IsodoseTab.py
Fincap: That's an interesting and powerful function, thanks for bringing it to my attention.

PR #59, opened 2020-10-27
Clinical data tab

Fincap:
This re-adds the clinical data tab that went missing during the main window restructures. Due to the sheer massive size of the two classes, there isn't enough time to fully refactor the ClinicalDataDisplay and ClinicalDataForm before the code freeze. The clinical data tab has the full functionality that it previously had, and additionally a bug that would cause a crash on Windows systems when saving new clinical data has now been fixed.


src/View/mainpage/MainPage.py
```diff
@@ -94,6 +96,23 @@ def setup_ui(self, main_window_instance):
...
+        # TODO refactor the entire Clinical Data form/display class
+        # As they currently stand, they are given the right tab widget, and make direct modifications to the tab.
+        # This class should be refactored in the same way as the rest of the main window's components, i.e. the Clinical
+        # Data should be a child of QWidget that can exist independently of OnkoDICOM. This class would differ from most
+        # other main window components in that rather than interacting with the PatientDictContainer as its model, it
+        # would use the patient's ClinicalData csv file as the model (which means that the QWidget would theoretically
+        # easily exist outside OnkoDICOM).
+        # There are two classes: one for displaying the clinical data, and another for modifying the clinical data.
+        # The check below determines whether there already exists a clinical data csv for the patient, and loads either
+        # the data display or the data form depending on what exists.
```
Fincap, line 109: "Vision statement" for the refactor of the Clinical Data tab. As it stands, the component isn't bad, it's just inconsistent with the other main window components.

PR #60, opened 2020-10-28
Adding more ROI Draw functionality

parise22:


requirements.txt
```diff
@@ -1,13 +1,16 @@
+numpy
```
sjswerdloff, line 1: the version constraint for numpy has been lost.  Is that intentional?

src/Model/ROI.py
```diff
@@ -1,6 +1,12 @@
 import collections
+import random
+
+import pydicom
+from pydicom import Sequence, Dataset
+from pydicom.tag import Tag
 
 from src.Model.CalculateImages import *
```
sjswerdloff, line 8: prefer explicit import rather than *

src/Model/ROI.py
```diff
@@ -46,6 +52,105 @@ def delete_roi(rtss, roi_name):
...
+        referenced_sop_instance_uid = data_set.ReferencedImageSequence[0].ReferencedSOPInstanceUID
+
+    referenced_frame_of_reference_uid = rtss["StructureSetROISequence"].value[0].ReferencedFrameOfReferenceUID
+    roi_number = rtss["StructureSetROISequence"].value[-1].ROINumber+1
+
+    # Colour TBC
+    red = random.randint(0,255)
+    green = random.randint(0,255)
+    blue = random.randint(0,255)
+    rgb = [red, green, blue]
```
sjswerdloff, line 82: even though this is randomized, there is a risk of having the same random colour used for multiple ROIs unless one makes sure that the particular rgb isn't used multiple times.

sjswerdloff (approved):
confirm that numpy version constraint isn't an issue before merging

PR #61, opened 2020-10-29
Improvements to saving modified RTSTRUCT.

Fincap:
This branch introduces the following:
- Saving RTSTRUCT no longer asks for a user file path. Instead, it now overwrites the existing RTSTRUCT file. The user is warned of this when they save.
- Fixed a bug that made multiple modified indicators appear on top of eachother when making multiple changes to the structures.
- The structure modified indicator now appears above the ROI Draw/Delete buttons.

### DIFF
--- a/src/View/mainpage/StructureTab.py
+++ b/src/View/mainpage/StructureTab.py
@@ -256,7 +256,12 @@ class StructureTab(QtWidgets.QWidget):
 
         self.modified_indicator_widget.setLayout(modified_indicator_layout)
         self.modified_indicator_widget.mouseReleaseEvent = self.save_new_rtss  # When the widget is clicked, save the rtss
+
+        # Temporarily remove the ROI modify buttons, add this indicator, then add them back again.
+        # This ensure that the modifier appears above the ROI modify buttons.
+        self.structure_tab_layout.removeWidget(self.roi_buttons)
         self.structure_tab_layout.addWidget(self.modified_indicator_widget)
+        self.structure_tab_layout.addWidget(self.roi_buttons)
 
     def structure_checked(self, state, roi_id):
         """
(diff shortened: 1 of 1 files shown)


lw803 (approved):
-Great

src/View/mainpage/StructureTab.py
```diff
@@ -279,10 +284,15 @@ def structure_checked(self, state, roi_id):
...
-        rtss_directory = str(Path(self.patient_dict_container.get("file_rtss")).parent)
-        save_filepath = save_filepath = QtWidgets.QFileDialog.getSaveFileName(self.parentWidget(), "Save file",
-                                                                              rtss_directory)[0]
-        if save_filepath != "":
-            self.patient_dict_container.get("dataset_rtss").save_as(save_filepath)
+        rtss_directory = str(Path(self.patient_dict_container.get("file_rtss")))
+
+        confirm_save = QtWidgets.QMessageBox.information(self, "Confirmation",
+                                                 "Are you sure you want to save the modified RTSTRUCT file? This will "
+                                                 "overwrite the existing file. This is not reversible.",
```
sjswerdloff, line 291: you might want to back up the original somewhere that won't get automatically loaded in a future run, but a support person could get to (e.g. rtss.dcm -> rtss.dcm.orig, and then you overwrite rtss.dcm... and if there is already an rtss.dcm.orig, start incrementing rtss.dcm.orig.001 or don't bother overwriting again... they have the true original and incremental changes are not as important... but if this is where the RadOnc is putting all of their effort, they might not be happy to lose incremental work if they made some kind of mistake).

Fincap: I will create this into a GitHub issue.

PR #72, opened 2020-11-06
Add ref img seq on create / prepopulate rtss

sjswerdloff:
The ROI module is looking for the ReferencedImageSequence in create_roi
The pre-populated rtss did not have that set (only the Contour Image Sequence was set)


kylelw23:
Just tested if there is still the crash that Andrew said in Slack, and it works

PR #74, opened 2020-11-09
Added save action

Fincap:
When an RTSTRUCT is present, a new Save Structure Changes button has been added to the toolbar and menubar. This is an alternative way for the user to save structures beyond the current options of clicking the modified structures indicator, or closing the current patient session.


kylelw23:
Tested

PR #75, opened 2020-11-09
User is now informed when DVH values are outdated after ROI draw.

Fincap:
When adding a contour to an existing ROI, or creating a new region entirely,the DVH graph will not reflect this changes. Until the functionality to re-calculate the DVH is implemented, the user will for the time being be informed that the modified DVHs are not up to date with the calculation.


kylelw23:
Tested:
image

PR #78, opened 2020-11-17
Improvements to multi-threading Worker class

Fincap:
R#1020
The Worker class has been used as a portable way of deploying and running any function in a new PyQt thread utilizing the Qt's QThreadPool and QRunnable classes, and connecting to the thread's progress/results/errors as PyQt signals.

In the previous implementation of the Worker class, any function called by the Worker class would require the called function to have a `progress_callback` parameter, representing the progress signal of the Worker class. This change makes it so that any function can be run with or without the parameter. If the Worker instance wants to emit a progress signal, it can do so by doing the following:
* In the instantiation of the worker class, set the keyword argument `progress_callback=True`. For example:
`worker = Worker(slow_calc_fn, list_of_inputs, progress_callback=True)`
* Add the parameter `progress_callback` to the declaration of the function being called, and then emit progress changes to the callback argument. For example:
```
def slow_calc_fn(list_of_inputs, progress_callback):
    progress = do_part_of_calc(...)
    progress_callback.emit(progress)
```
* In the scope of the new Worker class instance, connect to the instance's progress signal via:
`worker.signals.progress.connect(self.update_progress_bar)`

This is all explained within the Worker class' declaration in the new version. This merge also improves upon the documentation within the class, providing more specific details on how the class is used/how it works.

### DIFF
--- a/src/Model/Worker.py
+++ b/src/Model/Worker.py
@@ -33,7 +33,8 @@ class Worker(QRunnable):
         self.kwargs = kwargs
         self.signals = WorkerSignals()
 
-        self.kwargs['progress_callback'] = self.signals.progress
+        if 'progress_callback' in self.kwargs and self.kwargs['progress_callback'] is True:
+            self.kwargs['progress_callback'] = self.signals.progress
 
     @pyqtSlot()
     def run(self):
--- a/src/Model/Worker.py
+++ b/src/Model/Worker.py
@@ -14,6 +14,8 @@ class WorkerSignals(QObject):
 class Worker(QRunnable):
     """
     Multi-threading worker class that is executed by a QThreadPool.
+    The purpose of this class is to provide a simple way of executing any function on a separate thread, which proves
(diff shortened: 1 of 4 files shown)


sjswerdloff (approved):

src/Model/Worker.py
```diff
@@ -33,7 +35,24 @@ def __init__(self, fn, *args, **kwargs):
         self.kwargs = kwargs
         self.signals = WorkerSignals()
 
-        self.kwargs['progress_callback'] = self.signals.progress
+        # When the constructor for this class is called with a keyword argument 'progress_callback' set to True, this
```
sjswerdloff, line 38: Is there an automated test for this class?
Is it feasible to put together a small test (even calculating pi to ten decimal places would be fine), and maybe try it three different ways ( progress_callback is True, progress_callback is False, and no progress_callback kwarg parameter provided)?

Let's aim to add a minimal test for every module touched that doesn't have a test.
And if the module has a test, update or add to the test so the change itself is tested.

PR #80, opened 2020-11-24
Worker tests

Fincap:
Adds in unit testing for the Worker class to ensure that it will be instantiated properly, that the correct values are returned from the worker, and that exceptions are properly handled.

Also adds in a fix where the program would crash if progress_callback kwarg were set to False.

Added explanations of variables in exception handling part of Worker class.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ matplotlib==3.2.0
 PyWavelets==0.5.2
 dataclasses==0.7;python_version=="3.6.*"
 networkx
+pytest-qt
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,3 @@ matplotlib==3.2.0
 PyWavelets==0.5.2
 dataclasses==0.7;python_version=="3.6.*"
 networkx
-pytest-qt
--- a/src/Model/Worker.py
+++ b/src/Model/Worker.py
@@ -1,3 +1,4 @@
+import logging
 import sys
 import traceback
 
@@ -66,7 +67,7 @@ class Worker(QRunnable):
             result = self.fn(*self.args, **self.kwargs)
         except:
             # Emit signal error when exception occurs
-            traceback.print_exc()
+            logging.exception(traceback.format_exc())
             exctype, value = sys.exc_info()[:2]
             self.signals.error.emit((exctype, value, traceback.format_exc()))
         else:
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -7,12 +7,12 @@ scikit-build
...
 country-list
 git+https://github.com/matplotlib/legacycontour.git
 fuzzywuzzy
 matplotlib==3.2.0
 PyWavelets==0.5.2
 dataclasses==0.7;python_version=="3.6.*"
-networkx
\ No newline at end of file
+networkx
+pytest-qt
```
sjswerdloff, line 18: Is this something every user needs to have, or just developers.
It isn't unusual to have testing related packages dealt with separately (in a developer oriented requirements-dev.txt or something like that).

src/Model/Worker.py
```diff
@@ -53,6 +54,8 @@ def __init__(self, fn, *args, **kwargs):
 
         if 'progress_callback' in self.kwargs and self.kwargs['progress_callback'] is True:
```
sjswerdloff, line 55: blah is True
can more efficiently be left as
blah
`if 'progress_callback' in self.kwargs and self.kwargs['progress_callback']:`
```suggestion
        if 'progress_callback' in self.kwargs and self.kwargs['progress_callback']:
```

src/Model/Worker.py
```diff
@@ -53,6 +54,8 @@ def __init__(self, fn, *args, **kwargs):
...
             self.kwargs['progress_callback'] = self.signals.progress
+        elif 'progress_callback' in self.kwargs and self.kwargs['progress_callback'] is False:
```
sjswerdloff, line 57: elif 'progress_callback' in self.kwargs and not `self.kwargs['progress_callback']`
```suggestion
        elif 'progress_callback' in self.kwargs and not self.kwargs['progress_callback']:
```

sjswerdloff (changes requested):
please confirm or refute that the more concise conditional expressions work correctly (implement if they do).
remove pytest-qt from requirements if that's just for developers (but make sure that it gets installed as part of the automated testing!).

sjswerdloff (approved):

sjswerdloff:
test/test_model_anon.py ...                                               [ 30%]
test/test_model_roi.py ..                                                    [ 50%]
test/test_model_worker.py .....                                          [100%]

Happy thing!

src/Model/Worker.py
sjswerdloff: no need...
this is why we have code review, and this in particular was a nit

PR #83, opened 2020-12-02
skip attempt at pseudonymising PatientWeight at least until pymedphys…

sjswerdloff:
… fixes the bug involving DS


sjswerdloff:
Redmine Bug #1030

PR #88, opened 2020-12-17
Improved feedback when modifying RTSTRUCT

Fincap:
R#1021
These changes show the user visual feedback when Drawing, Deleting, or Renaming ROIs. Previously, the window would freeze while the RTSTRUCT would be updated, however that is unideal and can lead to users believing the program is no longer working. Now, a small window will pop up advising the user that some process is being run in the background and to wait for its completion.

### DIFF
--- a/src/Model/ROI.py
+++ b/src/Model/ROI.py
@@ -31,9 +31,12 @@ def delete_list_of_rois(rtss, rois_to_delete):
     Call the delete_roi function for each ROI in the given list.
     :param rtss: Dataset of RTSS.
     :param rois_to_delete: List of ROI names.
+    :return: Updated RTSS with deleted ROIs.
     """
     for item in rois_to_delete:
-        delete_roi(rtss, item)
+        rtss = delete_roi(rtss, item)
+
+    return rtss
 
 
 def delete_roi(rtss, roi_name):
--- a/src/Model/ROI.py
+++ b/src/Model/ROI.py
@@ -26,6 +26,16 @@ def rename_roi(rtss, roi_id, new_name):
     return rtss
 
 
+def delete_list_of_rois(rtss, rois_to_delete):
+    """
+    Call the delete_roi function for each ROI in the given list.
+    :param rtss: Dataset of RTSS.
+    :param rois_to_delete: List of ROI names.
+    """
+    for item in rois_to_delete:
+        delete_roi(rtss, item)
+
+
 def delete_roi(rtss, roi_name):
     """
     Delete ROI by name
--- a/src/View/mainpage/DeleteROIWindow.py
+++ b/src/View/mainpage/DeleteROIWindow.py
@@ -307,11 +307,15 @@ class UIDeleteROIWindow():
                                                       'Would you like to continue?',
                                                       QMessageBox.Yes | QMessageBox.No)
         if confirmation_dialog == QMessageBox.Yes:
-            for item in self.regions_of_interest_to_delete:
-                new_dataset = ROI.delete_roi(self.dataset_rtss, item)
+            progress_window = DeleteROIProgressWindow(self, QtCore.Qt.WindowTitleHint)
+            progress_window.signal_roi_deleted.connect(self.on_rois_deleted)
+            progress_window.start_deleting(self.dataset_rtss, self.regions_of_interest_to_delete)
+            progress_window.show()
 
-            self.deleting_rois_structure_tuple.emit((new_dataset, {"delete": self.regions_of_interest_to_delete}))
-            self.close()
+    def on_rois_deleted(self, new_rtss):
+        self.deleting_rois_structure_tuple.emit((new_rtss, {"delete": self.regions_of_interest_to_delete}))
+        QMessageBox.about(self, "Saved", "Regions of interest successfully deleted!")
+        self.close()
 
 
 class DeleteROIProgressWindow(QtWidgets.QDialog):
--- a/src/View/mainpage/DeleteROIWindow.py
+++ b/src/View/mainpage/DeleteROIWindow.py
@@ -1,4 +1,5 @@
-from PyQt5 import QtCore, QtGui
+import pydicom
+from PyQt5 import QtCore, QtGui, QtWidgets
 from PyQt5.QtCore import Qt
 from PyQt5.QtGui import QIcon, QPixmap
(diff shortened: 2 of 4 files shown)


src/Model/ROI.py
```diff
@@ -26,6 +26,19 @@ def rename_roi(rtss, roi_id, new_name):
     return rtss
 
 
+def delete_list_of_rois(rtss, rois_to_delete):
```
sjswerdloff, line 29:
```suggestion
def delete_list_of_rois(rtss : Dataset, rois_to_delete : List[str]):
```

while typing is loose in Python, providing it in signatures is useful  
you'll also need to add

from typing import List

src/Model/ROI.py
```diff
@@ -26,6 +26,19 @@ def rename_roi(rtss, roi_id, new_name):
...
+    """
+    Call the delete_roi function for each ROI in the given list.
+    :param rtss: Dataset of RTSS.
+    :param rois_to_delete: List of ROI names.
+    :return: Updated RTSS with deleted ROIs.
+    """
+    for item in rois_to_delete:
```
sjswerdloff, line 36: item has a particular connotation in DICOM, i.e. a Sequence Item.
So that makes this a bit confusing.


```suggestion
    for roi_name in rois_to_delete:
```

src/Model/ROI.py
```diff
@@ -26,6 +26,19 @@ def rename_roi(rtss, roi_id, new_name):
...
+        rtss = delete_roi(rtss, item)
```
sjswerdloff, line 37:
```suggestion
        rtss = delete_roi(rtss, roi_name)
```

src/View/mainpage/DeleteROIWindow.py
```diff
@@ -7,6 +8,8 @@
 from src.Model import ROI
 import platform
```
sjswerdloff, line 9: sort imports
(using isort)

src/View/mainpage/DeleteROIWindow.py
```diff
@@ -304,13 +307,50 @@ def confirm_button_onClicked(self):
...
-                new_dataset = ROI.delete_roi(self.dataset_rtss, item)
-
-            self.deleting_rois_structure_tuple.emit((new_dataset, {"delete": self.regions_of_interest_to_delete}))
-            self.close()
-
-
-
+            progress_window = DeleteROIProgressWindow(self, QtCore.Qt.WindowTitleHint)
+            progress_window.signal_roi_deleted.connect(self.on_rois_deleted)
+            progress_window.start_deleting(self.dataset_rtss, self.regions_of_interest_to_delete)
```
sjswerdloff, line 312: is the dialog modal?
if not, what happens if someone starts doing other things while waiting for the ROIs to be fully deleted?  Are the ROIs contained in something thread safe?

src/View/mainpage/DrawROIWindow.py
```diff
@@ -684,6 +692,46 @@ def set_selected_roi_name(self, roi_name):
         self.roi_name_line_edit.setText(self.ROI_name)
 
 
+class SaveROIProgressWindow(QtWidgets.QDialog):
```
sjswerdloff, line 695: this seems to share a fair amount with DeleteROIProgressWindow
Would it make sense to have a base class and derive from it?
It might not be worth it, but once you have started to repeat yourself ("DRY"), it is worth evaluating.

src/View/mainpage/RenameROIWindow.py
```diff
@@ -101,18 +105,58 @@ def on_text_edited(self, text):
...
         self.close()
 
     def on_ROI_clicked(self):
         clicked_ROI = self.list_of_ROIs.currentItem()
         # Excluding headers from being clicked.
         if not str(clicked_ROI.text()).startswith("------------Standard"):
             self.input_field.setText(str(clicked_ROI.text()))
+
+
+class RenameROIProgressWindow(QtWidgets.QDialog):
```
sjswerdloff, line 126: OK.  Now that there are three classes that share a lot, it's likely that creating a base class and deriving would be worthwhile.

sjswerdloff (changes requested):
Excellent additions and improvement to the UX (User Experience).

Given that this isn't an urgent bug fix, rather than suggesting future improvement, I'm requesting changes...  Please evaluate each comment, and either make changes or substantiate why it's better to leave things as they are.

src/Model/ROI.py
Fincap: I've actually been using strong typing in my personal projects outside Onko because it makes it a lot clearer, but I felt it was better to have a consistent style in Onko than my preferred style.

I think a discussion could be had regarding updating the style guide in redmine to encourage strong typing (I would definitely be on the side supporting strong typing).

src/View/mainpage/DeleteROIWindow.py
Fincap: There is (in my opinion) a significant aesthetic need for consistency in imports across the board that isn't currently being met. In this class, for example, why are we importing QtWidgets _and_ 10 other classes from QtWidgets as well. In my opinion most files which depend on PyQt5 should import QtCore, QtGui, and QtWidgets as a baseline, and then call necessary classes from there. It's something I've been slowly trying to fix up as I work on files, but it does sometimes slip through the cracks. I think an entire branch should be dedicated to cleaning this up across the entire project.

src/View/mainpage/DeleteROIWindow.py
Fincap: It is modal, yes. The first parameter of the constructor (in this case `self`) is the parent window of the window.

src/View/mainpage/RenameROIWindow.py
Fincap: I'll spend a bit more time on this, because I think you're probably right.

src/View/mainpage/DeleteROIWindow.py
sjswerdloff: while I don't disagree with your statements,  my comment was more about having the import platform moved to the top section of imports (built-ins) and placed in alphabetical order within that section of imports.
the isort tool has a set of rules for ordering the imports so that they are easier to look through.

The rules regarding *how* to import, e.g. with respect to namespaces, etc. are important, and that's an issue worth raising, so I've done so.

PR #90, opened 2020-12-18
Rejig requirements

Fincap:
Due to issues with the way pip works, despite Cython being listed as a dependency for PyWavelets it does not install automatically. There is a similar issue with numpy. There was also an issue being caused by PyRadiomics not being installed without dependencies. For these reasons, the requirements have been restructured to have pre-requirements. These will be installed, and then the regular requirements will be installed. The GitHub Action for testing has been updated to accommodate this change.

Additionally, the README has been updated to provided better and more up-to-date information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
 <p align="center"><img src="src/res/images/onkodicom_main_banner.png?raw=true" alt="main-icon-onko-dicom" width="250"></p>
 
 # Onko
-OnkoDICOM was created with Radiation Oncologists to allow Radiation Oncologists to do research on DICOM standard image sets (DICOM-RT, CT, MRI, PET) using open source technologies, such as pydicom, dicompyler-core, Pyqt5, PIL, and matplotlib. OnkoDICOM is cross platform, open source software, and welcomes contributions from the wider community via GitHub https://github.com/didymo/OnkoDICOM.
+OnkoDICOM was created with Radiation Oncologists to allow Radiation Oncologists to do research on DICOM standard image
+sets (DICOM-RT, CT, MRI, PET) using open source technologies, such as pydicom, dicompyler-core, Pyqt5, PIL, and
+matplotlib. OnkoDICOM is cross platform, open source software, and welcomes contributions from the wider community via
+GitHub https://github.com/didymo/OnkoDICOM.
 
 
 OnkoDICOM was inspired by the [dicompyler project](https://github.com/bastula/dicompyler).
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -22,31 +25,36 @@ installed by running this command in the terminal.
...
+OR for Windows
 
 `venv/Scripts/activate.bat`
 
 Install the requirements
 
-`pip install -r requirements.txt`
-
-`pip install --no-deps -r requirements-without-deps.txt`
+`pip install -r pre-requirements.txt`
```
sjswerdloff, line 55: what version of pip were you using?
while I have seen some weird things with pip, this seems like there might be something going on that should either be in a bug report or a release note in those other projects

requirements.txt
```diff
@@ -1,8 +1,5 @@
 alphashape
-numpy~=1.19
 pandas
-cython
-cpython
```
sjswerdloff, line 5: you moved numpy and cython in to pre-requirements.txt
but not cpython.
any specific reason for not having that in pre-requirements.txt?

README.md
```diff
@@ -22,31 +25,36 @@ installed by running this command in the terminal.
...
+If you are installing onto Windows, you will need to have installed Visual Studio Build Tools, and a 64-bit version of
+Python.
+
 Clone the Onko repository
 
 `git clone https://github.com/didymo/OnkoDICOM.git`
 
 Enter the directory and create a virtual environment with a name of
-your choice, in this case it's envOnkoDICOM.
+your choice, in this case it's venv.
```
sjswerdloff, line 36: if you are going to change this, it might be best to put the virtual env outside of the cloned repo.
I didn't really know too much about what I was doing with the virtual env at the time, but in retrospect, I should have put it outside of the repo so it wouldn't show up when doing things with git.

sjswerdloff (approved):
but please look at the comments first

README.md
Fincap: The base version of pip that is created with the virtual environment is adequate. The issue relating to the installed pip version is related to the way that specifically PyCharm installs pip into the virtual environment, but if the environment is created from the command line outside of PyCharm, it works correctly. This is a widely experienced issue with PyCharm apparently, e.g. https://github.com/pypa/pip/issues/5343

requirements.txt
Fincap: I believe CPython was incorrectly included in requirements to begin with. CPython is an implementation of the python language, whereas Cython is a superset to the Python language. I believe somewhere someone has accidentally included CPython as a requirement instead of Cython by mistake. It is not depended on by Onko or any of its requirements. The CPython package on pypi appears to have nothing to do with "real" CPython.

PR #93, opened 2021-03-24
Remove 'legacy' directory

Fincap:
A long overdue change to remove the legacy directory. The files that existed in the directory were no longer being used. The existence of the directory contradicts the purpose of using version control.


sjswerdloff (approved):
Hurrah 😊

PR #97, opened 2021-05-21
Comment explaining conversion "odd" of pixel array type.

Fincap:


Regnum771:
The comment explains the purpose for converting np_pixels to np.int16. It improves code readability and future-proof the case where someone removes this without knowing what it does.

Alamancy:
The comment provides context for why np_pixels needs to be type cast to np.int16 and allows for people reviewing the code to easily understand the purpose of the conversion.

sjswerdloff (commented):
Presumably the int16 data type is  used because some function expects it, and the reason why the function was designed to operate on data of that type was because that data type is returned by a well known third party utility function which is either the most common source of the data (pragmatic choice) for this application or library, or it was the first data source (legacy)
Or, the data is going to be fed back eventually to a third party library /package that expects the data to be in that format.

So what is the underlying reason?
Feeding to application functions that need the data this way?  Or to a third party package? 

Another way to look at this is 'why shouldn't single precision float be used?'
Or made as an option (e.g. by passing in an optional parameter indicating the desired return type)?

Fincap:
@sjswerdloff Having had a scour through the DICOM standard and PyDicom's source, I have found that the dtype of the numpy array returned by convert_pixel_data() depends on two Data Elements: `(0028,0101) BitsAllocated` and `(0028,0103) PixelRepresentation`. The [numpy handler function](https://github.com/pydicom/pydicom/blob/c73451668d9d2bf021bbec4067e31fb25083d582/pydicom/pixel_data_handlers/util.py#L1100) specifically returns the array in the same type as is defined by the aforementioned data elements. As a conjecture, it appears as though Onko currently is only designed and tested with uint16 pixel data in mind, however it is possible for DICOM pixel data to be signed or unsigned, and 1, 8, 16, 32, or 64 bits per value. You're right that this should be accommodated for.

PR #101, opened 2021-06-27
Move 'res' and 'data' directories out of 'src'

Fincap:
This change reorganises the project's directory structure and pulls the 'res' and 'data' directory that are currently contained in 'src' into the project's root directory. This makes the project's structure more clear and creates a more distinct separation between the project's source code (contained in 'src') and the resources it uses ('res' and 'data').

Also contained in this branch is a modification to the Windows .spec file for .exe compilation. The output directory no longer contains the project's source code, and an improvement has been made to the way Pyinstaller locates .dll files.


DuyTungHa:
I think this is very good. We should definitely restructure the project in this way.

src/View/FontsService.py
```diff
@@ -15,7 +15,7 @@ def __init__(self):
 		if(current_os == 'Windows'):
 			path_to_fonts += '\\src\\res\\fonts\\'
```
sjswerdloff, line 16: Is this a missed reference still pointing to src?

sjswerdloff (approved):
Consider creating a central definition of the directories to make future moves or local copies for customization less intensive

src/View/FontsService.py
Fincap: Good catch. I tried to a variety of combinations of paths to find uses of the paths. Will do another sweep through.

Fincap:
Forgive the poor regex, but I searched the entire project directory with the pattern `src.+(res)|src.+(data)` and I believe I've tracked down all filepath values that need to be changed.

didymo:
Better directory organisation.

PR #102, opened 2021-06-28
Pyside6 Integration

DuyTungHa:
Note: The matplotlib version is set to a pull request which will be merged and officially released in v.3.5.0.


Fincap (commented):
While the scope of this branch is quite clear and focused, I would recommend trying to put as much detail in the pull request description regarding the changes made. Obviously a file-by-file list of each module updated to PySide6 would be excessive, but I noticed a couple extra changes in there that aren't mentioned in the description.

Additionally, while I know you said it was a temporary solution for testing, I would seriously recommend against attaching an entire DICOM dataset to the project, even if it's been anonymised. This increases the repository size from somewhere under 10 MB to over 100 MB. if it's necessary to make unit testing work, is an entire dataset necessary? Is there any reason why you can't have a single image slice, a single RTSTRUCT, a single RTDOSE, etc.

PR #105, opened 2021-06-30
Addition of structures tab unit test

Tamookk:
These commits add a unit test for checking that changes to the structure tab's checkboxes result in the appropriate changes in the DICOM view tab. These commits assume the presence of sample DICOM files inside the folder `test/testdata/DICOM-RT-TEST/`.


DuyTungHa (commented):
1. The tests are good. No error occurred during the execution of the test. It covers for both cases of checking and unchecking item in the structure tab. 
2. Because pyside6-integration has been merged to master, this PR should go to master instead of pyside6-integration.
3. Could you rename the test file to test_view_structures_tab.py ? It follows the naming convention in other test files.

Tamookk:
I have renamed the test file to follow the naming convention and changed the base branch to the master branch.

didymo:
https://www.python.org/dev/peps/pep-0257/

check the commenting

Tamookk:
Docstrings have been added to the functions and class.

sjswerdloff (commented):
I enabled the workflow to be run, and am waiting to see how that turns out.  The dependency on test data means that either the test data has to be loaded as part of the workflow prior to the test, or the test has to check for the presence of the data and load it if not present (and potentially remove it when done, depending on how the loading of the test data was done).  This aspect of CI is challenging.

sjswerdloff:
The new test passes without the test data and without a mention of the test data not being present.

Not specific to this PR, but evident from careful review of the output of the workflow / Action... 

There are many depreciation warnings (I will try to attend to some of them). 

There are a huge number of flake8 warnings, and it would be nice to slowly work through them, preferably in PRs that do nothing else but deal with reformatting to meet the flake8 criteria.

Tamookk:
I can change the unit test to fail without test data present if you feel that would be necessary. As for the warnings, I also noticed a lot of them. Perhaps another aim for this year could be added to work on fixing these?

sjswerdloff:
For now, don't fail.
But please try to emit some text to indicate that in essence the test was skipped.
We need to build data download infrastructure for testing or else the data driven tests will all fail in CI.

sjswerdloff:
And yes, I think that setting aside a little time for cleanup would be good but you students are getting graded on other metrics, and we don't want you penalised for working on something that you don't get recognised for...

I think that a lot of this can be dealt with by only doing a clean up of a file you are going to touch.  Before changing code, reformat (only reformat) so a quick review is possible (if we had extensive unit tests, just passing after reformatting would be enough).

PR #106, opened 2021-06-30
Fixed bug with hard-coding indexes and missing rt files

JThomas2020:
This commit fixes an error when the program is missing rt files. The old code assumed the rt files were all included in loading and hard coded the indexes in. This code removes the hard coding and adds an array to track which rt files are present.


YukiTakeya256:
These commits fix a bug where opening an RTDose in the tree view when an RTPlan hasn't been loaded (or vice versa) crashes the program. They also fix the number of image slides that can be viewed in the tree view to accurately reflect the number of images. With the commits the program runs correctly and errors no longer occur.

PR #110, opened 2021-06-30
Bug Fix: Memory Leak in QMainWindow

DuyTungHa:
After a patient has been opened in the main page, if the user chooses to select another patient and go back to the open patient window, another instance of open patient window will be initialized in the program. 
This also happens to any QMainWindow that can be opened and closed multiple times. (Main Window and AddOnOptions)
**Solution is to prevent multiple initialization of window.**


Fincap (approved):
Looks good. Be mindful of reusing window instances as memory may not be cleaned up properly on reuse, however if this is tested to be memory safe then I think the change is good. The project has had issues in the past with improper reuse of window instances, so please be vigilant when making these kinds of changes, and thoroughly test for memory leaks.

didymo:
Demonstrate this is memory safe.

Fincap (changes requested):
I've pulled the changes down and tested them. There is a significant memory leak occurring when successive patients are opened. To see for yourself, simply open patient after patient and track the memory usage of the program using something like Task Manager. Not sure if this particular change has resulted in this or another recent change, but this should be urgently addressed.

DuyTungHa:
The open_patient_window and open_main_window functions always create a new instance of each of the windows when called. The old instance will be left in memory as a memory leak. Initializing open_patient_window and main_window only once resolves this issue. Could you please confirm this ? @Fincap

didymo:
this PR is undergoing further testing.

Fincap:
Upon further testing, it's memory usage appears to be consistent with previous builds. There is a slight leak of a couple megabytes every time a new patient is loaded, however the leak of hundreds of megabytes per patient appears to have been fixed. My testing method is quite barebones, but it help identify obvious and large leaks. I simply monitor the memory usage in Windows Task Manager, and repeatedly open and re-open the same DICOM dataset.

Here are my test results:

```
Dataset Size: 140MB
DVH: Not calculated

Initial startup: 	141.0MB
Patient loaded:		611.3MB

Open new patient 1: 	608.5MB
Patient loaded 1: 	726.2MB

Open new patient 2:	610.2MB
Patient loaded 2:	727.2MB

Open new patient 3:	611.0MB
Patient loaded 3:	727.6MB

Open new patient 4:	611.5MB
Patient loaded 4:	728.3MB

Open new patient 5:	612.5MB
Patient loaded 5:	728.9MB
```

Fincap (approved):

PR #113, opened 2021-07-06
add test for dvh tab

tdlinh2712:
this PR adds a unit test for DVH tab in OnkoDICOM.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,0 +1,23 @@
+import os
+import sqlite3
+
+import pytest
+
+from src.Model.Configuration import Configuration
+
+
+@pytest.fixture(scope="module", autouse=True)
+def init_config(request):
+    configuration = Configuration('TestConfig.db')
+    db_file_path = os.environ['USER_HIDDEN'] + 'TestConfig.db'
+    configuration.set_db_file_path(db_file_path)
+    connection = sqlite3.connect(db_file_path)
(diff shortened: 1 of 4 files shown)


didymo (approved):

DuyTungHa (changes requested):
The test runs forever in github pipeline. 
I think this is because the test does not account for configuration model. 
Please merge with master and make sure that all tests pass.

sjswerdloff:
> The test runs forever in github pipeline.
> I think this is because the test does not account for configuration model.
> Please merge with master and make sure that all tests pass.

Are you asking for the author to merge master back in to their branch?
Or to merge the PR in to master?
The latter should not happen if the test is failing (including running forever) and there isn't a clear explanation why that isn't relevant.

DuyTungHa:
@sjswerdloff I'm telling the author to pull from master and merge into their branch. This is because when I ran the test on my computer I got an error caused by a newly added feature in onkoDICOM. Probably the test hasn't accounted for this new feature.

tdlinh2712:
> The test runs forever in github pipeline.
> I think this is because the test does not account for configuration model.
> Please merge with master and make sure that all tests pass.

I have moved the init_config fixture outside of different files so that they can be used among files, which helped avoid the configuration issue.

DuyTungHa (approved):

PR #115, opened 2021-07-07
Renamed slots and method names

LWMcClymont:
Changed names of slots in file AddOnOptionsController to work with PySide's slot signature to avoid the following errors received on Linux: 

qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_customContextMenuRequested_Window(QPoint)

qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_customContextMenuRequested_Organ(QPoint)

qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_customContextMenuRequested_Volume(QPoint) 

Also changed method names to align with the naming conventions.

### DIFF
--- a/src/Controller/AddOnOptionsController.py
+++ b/src/Controller/AddOnOptionsController.py
@@ -97,7 +97,7 @@ class AddOnOptions(QtWidgets.QMainWindow, UIAddOnOptions):
         self.import_data(data)
         self.treeList.expandAll()
         # fill the corresponding tables with the corresponding data from the csv files
-        self.fillTables()
+        self.fill_tables()
         self.treeList.setEditTriggers(
             QtWidgets.QTreeView.NoEditTriggers
         )  # make the tree entries not editable
@@ -398,12 +398,12 @@ class AddOnOptions(QtWidgets.QMainWindow, UIAddOnOptions):
     #                                                                                                                 #
     ###################################################################################################################
 
-    def fillTables(self):
+    def fill_tables(self):
         # Fill the Windowing table
-        with open(resource_path("data/csv/imageWindowing.csv"), "r") as fileInput:
-            next(fileInput)
+        with open(resource_path("data/csv/imageWindowing.csv"), "r") as file_Input:
+            next(file_Input)
             i = 0
-            for row in fileInput:
+            for row in file_Input:
                 items = [
                     QTableWidgetItem(str(item.replace("\n", "")))
                     for item in row.split(",")
@@ -416,10 +416,10 @@ class AddOnOptions(QtWidgets.QMainWindow, UIAddOnOptions):
                 i += 1
 
         # organ names table
-        with open(resource_path("data/csv/organName.csv"), "r") as fileInput:
-            next(fileInput)
+        with open(resource_path("data/csv/organName.csv"), "r") as file_Input:
+            next(file_Input)
             i = 0
-            for row in fileInput:
+            for row in file_Input:
                 items = [
                     QTableWidgetItem(str(item.replace("\n", "")))
                     for item in row.split(",")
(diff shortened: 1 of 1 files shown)


src/Controller/AddOnOptionsController.py
```diff
@@ -398,12 +398,12 @@ def accepting(self):
     #                                                                                                                 #
     ###################################################################################################################
 
-    def fillTables(self):
+    def fill_tables(self):
         # Fill the Windowing table
-        with open(resource_path("data/csv/imageWindowing.csv"), "r") as fileInput:
-            next(fileInput)
+        with open(resource_path("data/csv/imageWindowing.csv"), "r") as file_Input:
```
sjswerdloff, line 403: if refactoring with renaming to meet python standards/defaults...
I believe that snake case for variables is appropriate,
file_input
rather than snake plus Camel Case...

sjswerdloff (approved):
While I'm approving, please consider the comment regarding snake case vs. snake + Camel case

src/Controller/AddOnOptionsController.py
```diff
@@ -147,7 +147,7 @@ def open_link(self, item):
     ######################################################################################################
     # windowing
     @Slot(QtCore.QPoint)
-    def on_customContextMenuRequested_Window(self, pos):
+    def on_customContextMenuRequestedWindow(self, pos):
```
sjswerdloff, line 150: But I don't see the change in signature other than the renaming, nor a change in annotation

src/Controller/AddOnOptionsController.py
```diff
@@ -116,15 +116,15 @@ def __init__(self, window):  # initialization function
         # adding the right click menus for each table
         self.table_view.setContextMenuPolicy(QtCore.Qt.ContextMenuPolicy.CustomContextMenu)
         self.table_view.customContextMenuRequested.connect(
-            self.on_customContextMenuRequested_Window
+            self.on_customContextMenuRequestedWindow
```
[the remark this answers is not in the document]
LWMcClymont: Although my understanding is still foggy, from what I can understand slots shouldn't be named in the format on_X_X as it conflicts with the default slot signature of on_<name_of_object>_<action>.

I should have worded the PR better, sorry for the confusion.

src/Controller/AddOnOptionsController.py
LWMcClymont: By renaming the slots (removing the last underscore) it solves the issue.
The next question would be is this appropriate? 
For example would onCustomContextMenuRequestedWindow be more appropriate than  on_customContextMenuRequestedWindow?

src/Controller/AddOnOptionsController.py
LWMcClymont: snake_case is correct, unfortunately I missed the capital "I". Thank you for pointing that out.

src/Controller/AddOnOptionsController.py
LWMcClymont: I'm still learning this - Any input you have to further my understanding is more than welcome!

sjswerdloff: The key thing is that the renaming is the fix.
I don't know anything about the signal and slot naming conventions.   I think Matthew is our local expert.

src/Controller/AddOnOptionsController.py
sjswerdloff: Please update with a commit or create a separate issue for follow up.
Prefer an update / commit because this code is exactly what was touched.

src/Controller/AddOnOptionsController.py
sjswerdloff: > I'm still learning this - Any input you have to further my understanding is more than welcome! 

That's what code review is good for.
We know that the teams are going through a learning curve and the biggest contribution I can make is helping individuals through the learning curve.

src/Controller/AddOnOptionsController.py
LWMcClymont: I've just pushed a commit up to fix this, thank you!

src/Controller/AddOnOptionsController.py
LWMcClymont: Thank you Stuart, it is well appreciated, and will continue to be appreciated throughout the year. Cheers!

sjswerdloff (approved):

PR #117, opened 2021-07-08
Addition of DICOM Tree View Tab Unit Test

JThomas2020:
This PR adds a unit test for the DICOM Tree View Tab in the main window. It checks the selector and the initial table have been loaded properly. Then for each file: the test recursively checks each row to determine if the data in the View corresponds to the matching data in PatientDictContainer. This commit assumes the existence of a test file in `/test/testdata/DICOM-RT-TEST`


JThomas2020:
Test works offline however having issues with loading PatientDictContainer, will re-open when sorted

JThomas2020:
Unit Test for DICOM Tree Tab has been refactored so PatientDictContainer is set up properly, unit test works as expected now. This pull request also has some PEP-8 refactoring, will make sure to separate these in the future. This test will be refactored in the future with qtbot support.

test/test_view_dicom_tree.py
```diff
@@ -0,0 +1,158 @@
...
+    return count
+
+
+class TestDICOMTreeTab:
+    """Class to set up the OnkoDICOM main window for testing the structures tab."""
+    __test__ = False
+
+    def __init__(self):
+        # Load test DICOM files and set path variable
+        path = "/testdata/DICOM-RT-TEST"
```
sjswerdloff, line 64: This anti pattern has become prevalent.
Use pathlib to avoid coding to specific platforms.

sjswerdloff (changes requested):

DuyTungHa (commented):
I have no comments other than that we should switch to pathlib as suggested by @sjswerdloff  
But maybe this should be done in a separate PR because there are other places in the application which use this kind of hard coded path. (e.g. stylesheet)

test/test_view_dicom_tree.py
didymo: I agree

didymo:
> I have no comments other than that we should switch to pathlib as suggested by @sjswerdloff
> But maybe this should be done in a separate PR because there are other places in the application which use this kind of hard coded path. (e.g. stylesheet)

It is a good idea to follow Stuart's suggestions when they are made and not later.

sjswerdloff:
> > I have no comments other than that we should switch to pathlib as suggested by @sjswerdloff
> > But maybe this should be done in a separate PR because there are other places in the application which use this kind of hard coded path. (e.g. stylesheet)
> 
> It is a good idea to follow Stuart's suggestions when they are made and not later.

I tend to be lenient about fixing stuff later.  I have mentioned this in other reviews.  So, no, this can't wait.  Someone has to be first, and it's better to have it be narrow instead of same change across ten files and a review comment that forces updates again in a zillion places.

If I don't make it clear that it is not urgent then it is at least somewhat urgent.  If I block a PR, then I am as serious as a heart attack that it needs to be dealt with. Ideally while you still have my attention and I'm not dealing with a crisis  at my day job.

test/test_view_dicom_tree.py
```diff
@@ -0,0 +1,157 @@
...
+
+class TestDICOMTreeTab:
+    """
+    Class to set up the OnkoDICOM main window for testing the structures tab.
+    """
+    __test__ = False
+
+    def __init__(self):
+        # Load test DICOM files and set path variable
+        path = str(Path.cwd() / 'test' / 'testdata' / 'DICOM-RT-TEST')
```
JThomas2020, line 66: code has been updated to use pathlib

sjswerdloff:
```suggestion
        path = Path.cwd().joinpath('test' , 'testdata' , 'DICOM-RT-TEST')
```

sjswerdloff (changes requested):
See suggestion

sjswerdloff (approved):

DuyTungHa (approved):

didymo (approved):

PR #118, opened 2021-07-09
Wrong image type in Open Patient Window

DuyTungHa:
The wrong image type in Open Patient Window prevents the application from opening MR scan. 
In relation to issue #1561 on Team23's Redmine.

### DIFF
--- a/src/View/OpenPatientWindow.py
+++ b/src/View/OpenPatientWindow.py
@@ -294,7 +294,7 @@ class UIOpenPatientWindow(object):
                 self.selected_series_types.update(series_type)
 
         # Check the existence of IMAGE, RTSTRUCT and RTDOSE files
-        if len(list({'CT', 'MRI', 'PET'} & self.selected_series_types)) == 0:
+        if len(list({'CT', 'MR', 'PET'} & self.selected_series_types)) == 0:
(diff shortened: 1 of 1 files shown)


sjswerdloff (approved):

Tamookk (approved):

JThomas2020 (approved):

src/View/OpenPatientWindow.py
```diff
@@ -294,7 +294,7 @@ def tree_item_changed(self, item, _):
                 self.selected_series_types.update(series_type)
 
         # Check the existence of IMAGE, RTSTRUCT and RTDOSE files
-        if len(list({'CT', 'MRI', 'PET'} & self.selected_series_types)) == 0:
+        if len(list({'CT', 'MR', 'PET'} & self.selected_series_types)) == 0:
```
sjswerdloff, line 297: Might be worthwhile to define different lists, including 'three_d_image_types', plan_types, or whatever it is that appears in more than one place.  That way you update the lists in one place.  Not critical nor urgent.
Technically, these are probably the values in the modality element in DICOM... so names of the lists might say modalities rather than types.

sjswerdloff (approved):

PR #119, opened 2021-07-09
Bug Fix: Update main window when a new patient is loaded

DuyTungHa:
The current main window does not update its UI when a new patient is opened. This PR fixes this problem. 
In relation to issue #1563 on Team23's Redmine.


sjswerdloff (approved):

Tamookk (approved):
Changes look good to me

JThomas2020 (approved):

PR #121, opened 2021-07-12
Bug#1564: ROI rename window not shown

tdlinh2712:
This PR addresses issue number #1564: OnkoDICOM malfunction - OnkoDICOM_Video2
The ROI rename function does not appear because in PySide6, functions named `exec_ `(classes QCoreApplication, QDialog, QEventLoop) have been renamed to `exec`.

### DIFF
--- a/src/View/mainpage/StructureWidget.py
+++ b/src/View/mainpage/StructureWidget.py
@@ -71,7 +71,6 @@ class StructureWidget(QtWidgets.QWidget):
         This function is called whenever the QWidget is right clicked.
         This creates a right click menu for the widget.
         """
-        print("here")
         # Part 1: Construct context menu
         menu = QtWidgets.QMenu(self)
         menu.setStyleSheet("QMenu::item::selected {background-color: #9370DB}")
@@ -87,7 +86,6 @@ class StructureWidget(QtWidgets.QWidget):
 
         # Part 2: Determine action taken
         action = menu.exec(self.mapToGlobal(event.pos()))
-        print(action)
         if action == rename_action:
             rename_window = RenameROIWindow(self.structure_tab.standard_volume_names,
                                             self.structure_tab.standard_organ_names,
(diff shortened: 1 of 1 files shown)


Tamookk:
If `menu.exec_` needs to be changed to `menu.exec`, then why has `rename_window.exec_` not been changed?

tdlinh2712:
> If `menu.exec_` needs to be changed to `menu.exec`, then why has `rename_window.exec_` not been changed?

From what I understand, It's because the function exec_() in class QMenu do not have the right arguments comparing to exec(), but in class QDialog exec_() and exec() are the same in terms of functionality.
Maybe I should change all exec_() to exec() for consistency?

Tamookk:
Thanks for the explanation. If `exec` and `exec_` in `QDialog` are the same then making these consistent would be a purely cosmetic change and could have unintended consequences.

src/View/mainpage/StructureWidget.py
```diff
@@ -55,7 +54,8 @@ def checkbox_clicked(self, checked):
 
     def roi_suggestions(self):
         """
-        Get the top 3 suggestions for the selected ROI based on string matching with standard ROIs provided in .csv format.
+        Get the top 3 suggestions for the selected ROI based on string matching with standard ROIs provided in .csv
```
fe-liii, line 57: To comply with PEP 8, comments and docstrings can only have a maximum of 72 characters per line.

DuyTungHa (approved):
It fixes the problem. 
Just need to follow some coding guidelines commented by @fe-liii.

src/View/mainpage/StructureWidget.py
didymo: Please address this.

tdlinh2712: done

didymo (approved):

PR #124, opened 2021-07-15
Product Backlog #1469: Addition of Isodoses Tab Unit Test

Tamookk:
This PR adds a unit test for the Isodoses tab in the main window. It asserts that the correct doses are selected and deselected, and therefore drawn and not drawn when the corresponding checkboxes in the Isodoses tab are checked and unchecked. This commit assumes the existence of a test file in /`test/testdata/DICOM-RT-TEST`.

### DIFF
--- a/test/test_view_isodoses_tab.py
+++ b/test/test_view_isodoses_tab.py
@@ -98,11 +98,10 @@ def test_isodoses_tab_check_checkboxes(test_object):
 
 
 def test_isodoses_tab_uncheck_checkboxes(test_object):
-    # For each available isolevel, simulate corresponding
-    # checkbox set to false
+    # For each available isolevel, simulate corresponding checkbox
+    # set to false
     for isolevel in isodose_percentages:
         test_object.main_window.isodoses_tab.checked_dose(False, isolevel)
         # Assert the isodose is not being drawn
         doses = test_object.patient_dict_container.get('selected_doses')
         assert isolevel not in doses
-
--- a/test/test_view_isodoses_tab.py
+++ b/test/test_view_isodoses_tab.py
@@ -1,22 +1,19 @@
 import os
-import platform
 import pytest
 from pathlib import Path
 
-from PySide6.QtWidgets import QApplication
 from src.Controller.GUIController import MainWindow
 from src.Model.PatientDictContainer import PatientDictContainer
 from src.View.mainpage.IsodoseTab import isodose_percentages
-from src.Model.ROI import get_contour_pixel
 from src.Model import ImageLoading
-import src.View.mainpage.DicomView
 
 from pydicom import dcmread
 from pydicom.errors import InvalidDicomError
 
 
 def find_DICOM_files(file_path):
-    """Function to find DICOM files in a given folder.
+    """
+    Function to find DICOM files in a given folder.
     :param file_path: File path of folder to search.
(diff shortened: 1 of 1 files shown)


test/test_view_isodoses_tab.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Class to set up the OnkoDICOM main window for testing the
+    structures tab.
+    """
+    __test__ = False
+
+    def __init__(self):
+        # Load test DICOM files
+        desired_path = str(Path.cwd().joinpath('test', 'testdata', \
+                                               'DICOM-RT-TEST'))
```
JThomas2020, line 46:
```suggestion
        desired_path = Path.cwd().joinpath('test', 'testdata', \
                                               'DICOM-RT-TEST')
```
str is not necessary for path

Regnum771 (approved):

test/test_view_isodoses_tab.py
```diff
@@ -0,0 +1,109 @@
...
+    for isolevel in isodose_percentages:
+        test_object.main_window.isodoses_tab.checked_dose(True, isolevel)
+        # Assert the isodose is being drawn
+        doses = test_object.patient_dict_container.get('selected_doses')
+        assert isolevel in doses
+
+
+def test_isodoses_tab_uncheck_checkboxes(test_object):
+    # For each available isolevel, simulate corresponding
+    # checkbox set to false
```
Regnum771, line 102:
```suggestion
    # For each available isolevel, simulate corresponding checkbox
    # set to false
```
Make the commenting consistent with its test_isodoses_tab_check_checkboxes test case and easier to read.

Tamookk:
I have committed both changes, thank you for the feedback

JThomas2020 (approved):

test/test_view_isodoses_tab.py
```diff
@@ -0,0 +1,107 @@
...
+    dicom_files = []
+
+    # Walk through directory
+    for root, dirs, files in os.walk(file_path, topdown=True):
+        for name in files:
+            # Attempt to open file as a DICOM file
+            try:
+                dcmread(os.path.join(root, name))
+            except (InvalidDicomError, FileNotFoundError):
+                pass
```
sjswerdloff, line 30: Given a known test input (set of files), wouldn't you know which files should fail?
It's arguable whether one should rethrow the exception if a failure is really not expected, but it at least implies something is wrong with the data.

Tamookk: It was written this way because the current test data set includes some non-DICOM files. I also personally believe that finding the test DICOM files programmatically is better than hard-coding their locations, as it future-proofs the test in the event that the test data changes in the future. The test dataset on my local machine is also larger than what is in this repository.

sjswerdloff: OK. Your test currently specifies a particular data set but you expect alternative data in the future.
My thinking was to identify known exceptions as a secondary filter after a failed read rather than identifying known good data explicitly.
Just wanted a documented discussion of the choices being made.  No wrong answer...

sjswerdloff (approved):

test/test_view_isodoses_tab.py
Tamookk: That's a good idea, I'll look into implementing that for any future tests I write that involve test data. Thank you for the feedback.

PR #126, opened 2021-07-22
Bug/fix open patient window bugs

minhhuyxx:
This PR fixes the problems described in tickets **#1605** and **#1580** on Team 23's Redmine page.

- When a new directory is scanned, the program disables the "Confirm" button in the Open patient window.
- If more than 1 study is selected, the program will disable the "Confirm" button and display a warning message in the DICOM tree's header.

### DIFF
--- a/src/View/OpenPatientWindow.py
+++ b/src/View/OpenPatientWindow.py
@@ -196,6 +196,7 @@ class UIOpenPatientWindow(object):
 
     def scan_directory_for_patient(self):
         # Reset tree view header and last patient
+        self.open_patient_window_confirm_button.setDisabled(True)
         self.open_patient_window_patients_tree.setHeaderLabels([""])
         self.last_patient = None
         self.filepath = self.open_patient_directory_input_box.text()


Tamookk (approved):

sjswerdloff:
While the IHE-RO profiles state that without a scheduled study instance UID the treatment planning related objects should reuse the Study Instance UID from the CT, what recourse will the user have if the data is not contained in one study but is in fact coherent?

The organising principle should rely on the hierarchy of the objects as defined by their references.  For example, if one selects a CT, only  RTSS that reference the CT should be allowed.  Once an RTSS is selected only RT Plans that reference the RTSS should be allowed.  Alternatively, if the RT dose gets selected one could directly/automatically select the referenced plan, then automatically select the referenced RTSS and then automatically select the set of referenced image series.

minhhuyxx:
Thanks, @sjswerdloff. The comment is very detailed and helpful. I will look into how to implement this and keep you updated.

tdlinh2712:
> While the IHE-RO profiles state that without a scheduled study instance UID the treatment planning related objects should reuse the Study Instance UID from the CT, what recourse will the user have if the data is not contained in one study but is in fact coherent?
> 
> The organising principle should rely on the hierarchy of the objects as defined by their references. For example, if one selects a CT, only RTSS that reference the CT should be allowed. Once an RTSS is selected only RT Plans that reference the RTSS should be allowed. Alternatively, if the RT dose gets selected one could directly/automatically select the referenced plan, then automatically select the referenced RTSS and then automatically select the set of referenced image series.

Hi Stuart, I agree with you on this comment, but that sounds like an extra feature that allows users to select different types of files without being restricted within a study and it will need further research, while this PR addresses a high-priority issue reported by Dr. Miller and that he wants multiple studies selection to be disabled for now. Therefore, I suggest merging this PR to resolve the issue first, and for the feature that you suggest, we will note that for further research and implementation.

tdlinh2712:
> > While the IHE-RO profiles state that without a scheduled study instance UID the treatment planning related objects should reuse the Study Instance UID from the CT, what recourse will the user have if the data is not contained in one study but is in fact coherent?
> > The organising principle should rely on the hierarchy of the objects as defined by their references. For example, if one selects a CT, only RTSS that reference the CT should be allowed. Once an RTSS is selected only RT Plans that reference the RTSS should be allowed. Alternatively, if the RT dose gets selected one could directly/automatically select the referenced plan, then automatically select the referenced RTSS and then automatically select the set of referenced image series.
> 
> Hi Stuart, I agree with you on this comment, but that sounds like an extra feature that allows users to select different types of files without being restricted within a study and it will need further research, while this PR addresses a high-priority issue reported by Dr. Miller and that he wants multiple studies selection to be disabled for now. Therefore, I suggest merging this PR to resolve the issue first, and for the feature that you suggest, we will note that for further research and implementation.

This suggestion has been documented in Redmine's Bug list (bug #1696).

PR #129, opened 2021-07-23
Bug Fix: Calculate DVHs Messagebox button ordering 

LWMcClymont:
Calculate DVHs Messagebox now displays "No" & "Yes" buttons in respective order across platforms.

### DIFF
--- a/src/View/ProgressWindow.py
+++ b/src/View/ProgressWindow.py
@@ -60,15 +60,16 @@ class ProgressWindow(QDialog):
         self.progress_bar.setValue(progress_update[1])
 
     def prompt_calc_dvh(self):
-        """ Windows displays buttons in a different order from Linux. A check for
-                    platform is performed to ensure consistency of button positioning across
-                    platforms.
-                """
+        """
+            Windows displays buttons in a different order from Linux. A check for
+            platform is performed to ensure consistency of button positioning across
+            platforms.
+        """
         if platform.system() == "Linux":
             choice = QMessageBox.question(self, "Calculate DVHs?", "RTSTRUCT and RTDOSE datasets identified. Would you "
-                                                                   "like to calculate DVHs? (This may take up to several "
-                                                                   "minutes on some systems.)",
-                                          QMessageBox.Yes | QMessageBox.No)
+                                                                   "like to calculate DVHs? (This may take up to "
+                                                                   "several minutes on some systems.)",
+                                                QMessageBox.Yes | QMessageBox.No)
 
             if choice == QMessageBox.Yes:
                 self.signal_advise_calc_dvh.emit(True)
--- a/src/View/ProgressWindow.py
+++ b/src/View/ProgressWindow.py
@@ -1,4 +1,4 @@
-import threading
+import threading, platform
 
 from PySide6 import QtCore, QtWidgets, QtGui
 from PySide6.QtCore import QThreadPool
@@ -7,6 +7,7 @@ from PySide6.QtWidgets import QDialog, QLabel, QVBoxLayout, QMessageBox
 from src.Model import ImageLoading
 from src.Model.Worker import Worker
 from src.View.ImageLoader import ImageLoader
+from src.Controller.PathHandler import resource_path
 
 
 class ProgressWindow(QDialog):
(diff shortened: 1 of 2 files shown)


src/View/ProgressWindow.py
```diff
@@ -59,15 +60,58 @@ def update_progress(self, progress_update):
...
+                respect to their assigned roles. (0 first, then 0 and so on) 'AcceptRole' is 0 and 'RejectRole' is 1 
+                thus by counterintuitively assigning 'No' to 'AcceptRole' and 'Yes' to 'RejectRole' the buttons are 
+                positioned as desired.
+            """
+            mb.addButton(button_no, QtWidgets.QMessageBox.AcceptRole)
+            mb.addButton(button_yes, QtWidgets.QMessageBox.RejectRole)
+
+            # Apply stylesheet to the message box and add icon to the window
+            mb.setStyleSheet(open(stylesheet_path).read())
+            mb.setWindowIcon(QtGui.QIcon(resource_path("res/images/btn-icons/onkodicom_icon.png")))
```
sjswerdloff, line 107: Is this expression of path platform independent?  If not, suggest using pathlib.

src/View/mainpage/DVHTab.py
```diff
@@ -133,14 +133,60 @@ def plot_dvh(self):
...
+                       "minutes on some systems.)")
+            button_no = QtWidgets.QPushButton("No")
+            button_yes = QtWidgets.QPushButton("Yes")
+
+            """ We want the buttons 'No' & 'Yes' to be displayed in that exact order. QMessageBox displays buttons in
+                respect to their assigned roles. (0 first, then 0 and so on) 'AcceptRole' is 0 and 'RejectRole' is 1 
+                thus by counterintuitively assigning 'No' to 'AcceptRole' and 'Yes' to 'RejectRole' the buttons are 
+                positioned as desired.
+            """
+            mb.addButton(button_no, QtWidgets.QMessageBox.AcceptRole)
```
sjswerdloff, line 176: That very similar code is being repeated suggests one might want to create a function that gets shared / reused.

src/View/ProgressWindow.py
LWMcClymont: I believe it was, it's been updated to use pathlib regardless.

src/View/mainpage/DVHTab.py
LWMcClymont: Yes, you are correct, however, I could not find a simple clean way of creating a single function that wouldn't require passing references from class to class which would get a little messy.
I suspect this is why the code was written the way it originally was.
 
My initial solution was to make the function static from within DVHTab that the other objects can reference, however, QMessageBox requires a reference to an instance of the parenting classes QWidget class. This cannot be accessed if the function were static.
From my understanding, this is where the styling of the buttons and such comes from. 
I could create a new QMessageBox object and assign the stylesheet as I have done for the windows correction, however on Linux versions, the style differentiates from what is currently shown on master for an unknown reason. 
I did not want to change too much of what is currently working. Please if you have any suggestions they are warmly welcome.

sjswerdloff: OK.  You examined the problem and there wasn't a reasonable solution that presented itself.  Fair enough!

src/View/ProgressWindow.py
```diff
@@ -59,15 +61,59 @@ def update_progress(self, progress_update):
...
+                self.signal_advise_calc_dvh.emit(True)
+            else:
+                self.signal_advise_calc_dvh.emit(False)
         else:
-            self.signal_advise_calc_dvh.emit(False)
+            stylesheet_path = ""
+
+            # Select appropriate style sheet
+            if platform.system() == 'Darwin':
+                stylesheet_path = "res/stylesheet.qss"
```
sjswerdloff, line 84: Not using pathlib here

sjswerdloff (approved):
Consider following through with pathlib on some resource paths.

src/View/ProgressWindow.py
LWMcClymont: Fixed - Thank you for spotting.

PR #131, opened 2021-07-25
Bug/Fix Draw ROI window

LWMcClymont:
This PR addresses the following issues on Linux systems when opening a Draw ROI window:

qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_zoom_out_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_backward_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_forward_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_box_draw_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_draw_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_preview_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_reset_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_cancel_button_clicked()
qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_save_clicked()

Please note whilst snake_case is the naming convention we follow, it appears to conflict with Qt, thus camelCase is used to fix the above errors.

### DIFF
--- a/src/View/mainpage/DeleteROIWindow.py
+++ b/src/View/mainpage/DeleteROIWindow.py
@@ -180,7 +180,7 @@ class UIDeleteROIWindow():
         self.delete_roi_window_cancel_button.resize(self.delete_roi_window_cancel_button.sizeHint().width(),
                                      self.delete_roi_window_cancel_button.sizeHint().height())
         self.delete_roi_window_cancel_button.setCursor(QtGui.QCursor(QtCore.Qt.PointingHandCursor))
-        self.delete_roi_window_cancel_button.clicked.connect(self.on_cancel_button_clicked)
+        self.delete_roi_window_cancel_button.clicked.connect(self.onCancelButtonClicked)
         self.delete_roi_window_cancel_button.setProperty("QPushButtonClass", "fail-button")
         self.delete_roi_window_action_buttons_box.addStretch(1)
         self.delete_roi_window_action_buttons_box.addWidget(self.delete_roi_window_cancel_button)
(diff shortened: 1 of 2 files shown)


didymo:
https://stackoverflow.com/questions/3647518/pep8-and-pyqt-how-to-reconcile-function-capitalization

LWMcClymont:
That is a good find. 

Adding the option `from __feature__ import snake_case, true_property` allows snake_case to be used for slots, however, it will break the rest of the file as all QT functions (which uses camelCase) will be expecting snake_case. 

The file can be refactored to accompany this, however the file will be inconsistent with the rest of the project unless willing to spend time refactoring the entire codebase. 

In this situation, I believe the best option is to not fight QT and use camelCase when declaring slot functions. It will be reliable.

LWMcClymont:
That is a good find. 

Adding the option `from __feature__ import snake_case, true_property` allows snake_case to be used for slots, however, it will break the rest of the file as all QT functions (which uses camelCase) will be expecting snake_case. 

The file can be refactored to accompany this, however the file will be inconsistent with the rest of the project unless willing to spend time refactoring the entire codebase. 

In this situation, I believe the best option is to not fight QT and use camelCase when declaring slot functions. It will be reliable.

didymo:
Agreed following along with the framework is always a good idea

tdlinh2712 (approved):

JThomas2020:
While converting all functions to snake_case is feasible, PEP-8 does not require QT to be written using snake_case and prefers consistency wih the QT library

From [https://www.python.org/dev/peps/pep-0008/#naming-conventions](url)

> New modules and packages (including third party frameworks) should be written to these standards, but where an existing library has a different style, internal consistency is preferred.

JThomas2020 (approved):

PR #133, opened 2021-07-27
GitHub CI Test Data

DuyTungHa:
This PR is in relation to Issue #1638 on Team23's Redmine Page. 
Now the CI process does not need the test data to be in the source code anymore. 
For local testing, you can still use the test/testdata directory. This directory is included in .gitignore file so it won't be pushed to the repository.


JThomas2020 (approved):
This requests pulls test data from the OnkoDICOM website and places it in a test folder. The other file changes fix the hardcoding path variables and changing the paths to access the files downloaded from the OnkoDIOCM website.

didymo (changes requested):
Do not pull anything from the Website. The website is likely to be changed without notice.

DuyTungHa:
> Do not pull anything from the Website. The website is likely to be changed without notice.

Do you recommend any other sources from which I can pull the test data?

sjswerdloff:
We need a zenodo group, placement of the test data on it, and then dynamic download of it.   See PyMedPhys tests and utilities for getting test data from zenodo

DuyTungHa:
@didymo Since you are the owner of the repository, could you go on zenodo and upload the DICOM zip files from onkoDICOM website there ? 
Here is an instruction video: https://youtu.be/S1qK_TA52e4 
Here is a sample group of testing data for PyMedPhys: https://zenodo.org/record/3870436#.YQIhYOgzaUk 
Once the test data are uploaded, I will change the github workflow to pull the test data from zenodo.

DuyTungHa:
@didymo I have updated the pipeline to pull the test data from https://zenodo.org/record/5147737#.YQOtO44zaUk using the links that you provided in slack.

didymo (changes requested):
Why is sudo being used for a user-space application?

DuyTungHa:
@didymo The first sudo is used to install the packages required to pull the test data from zenodo. The second sudo is used to modify the permission of each file in the test data because some files are marked as read-only after unzip and the application requires both read and write access. I have removed the unnecessary sudo in unzip.

didymo (changes requested):
Remove all operations requiring elevated permissions.
Meaning remove all sudo regardless of need.
Put these operations in the install wiki or another appropriate area.

DuyTungHa:
@didymo I'm not sure but I think you may have misunderstood the process of continuous integration on GitHub. An example will illustrate this process better. 
When I push my code to GitHub, GitHub will grab my pythonapp.yml and run all the operations within that file in a GitHub CI virtual machine. If everything runs successfully, I will see the "All checks have passed" in my pull request. This is to ensure that the code has passed all the unit test cases.
The actual user will not use or run anything inside pythonapp.yml at all. The user will follow the normal installation guide on wiki to install the packages and run the application.

didymo (approved):

PR #148, opened 2021-08-09
fix anonymized patient list can not display upon anonymisation

eddieGong2604:
After user clicks into Anonymisation button, a anonymized patient item is created but it is not updated to the anonymised patient list in add-on options


Tamookk (approved):
This PR fixes the issue where the anonymised patient list in add on options would not be updated after anonymising a patient. Team 14 has tested and approves these changes.

JThomas2020 (approved):
Team 19

PR #149, opened 2021-08-12
Fix anonymising saving to wrong directory

Tamookk:
When opening multiple patients, only the first patient that was opened would be anonymised, regardless of what was currently opened.

This was caused by the class `MainPageCallClass`, which wraps and controls the anonymisation function, not being updated whenever a new patient was open. Therefore, the file path and dataset being passed into the anonymisation function would always be the file path and dataset of the first patient that was opened.

This PR fixes this by creating a function to reload and update file path and dataset information in `MainPageCallClass`, and calling this function before anonymising the currently opened patient.

### DIFF
--- a/src/Controller/MainPageController.py
+++ b/src/Controller/MainPageController.py
@@ -1184,6 +1184,13 @@ class MainPageCallClass:
         self.dataset = self.patient_dict_container.dataset
         self.filepaths = self.patient_dict_container.filepaths
 
+    def reset(self):
+        """Resets class variables."""
+        self.patient_dict_container = PatientDictContainer()
+        self.path = self.patient_dict_container.path
+        self.dataset = self.patient_dict_container.dataset
+        self.filepaths = self.patient_dict_container.filepaths
+
     # This function runs Anonymization on button click
     def run_anonymization(self, raw_dvh):
         target_path = anonymize(self.path, self.dataset, self.filepaths,
--- a/src/Controller/MainPageController.py
+++ b/src/Controller/MainPageController.py
@@ -1180,33 +1180,29 @@ class MainPageCallClass:
     # Initialisation function of the controller
     def __init__(self):
         self.patient_dict_container = PatientDictContainer()
-        self.path = self.patient_dict_container.path
-        self.dataset = self.patient_dict_container.dataset
-        self.filepaths = self.patient_dict_container.filepaths
-
-    def reset(self):
-        """Resets class variables."""
-        self.patient_dict_container = PatientDictContainer()
-        self.path = self.patient_dict_container.path
-        self.dataset = self.patient_dict_container.dataset
-        self.filepaths = self.patient_dict_container.filepaths
 
     # This function runs Anonymization on button click
     def run_anonymization(self, raw_dvh):
-        target_path = anonymize(self.path, self.dataset, self.filepaths,
-                                raw_dvh)
+        path = self.patient_dict_container.path
+        dataset = self.patient_dict_container.dataset
+        filepaths = self.patient_dict_container.filepaths
+        target_path = anonymize(path, dataset, filepaths, raw_dvh)
         return target_path
 
     # This function displays the clinical data form
     def display_cd_form(self, tab_window, file_path):
-        self.tab_cd = ClinicalDataForm(
-            tab_window, file_path, self.dataset, self.filepaths)
+        dataset = self.patient_dict_container.dataset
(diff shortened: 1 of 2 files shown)


sjswerdloff (approved):
Given that the commit comment says it fixes saving to the wrong directory you might want to comment on what part of the change is responsible for achieving this.  I'm assuming that the reset of data paths is why the correct directory is now found

Tamookk:
Thank you for the feedback, I have updated the PR description to provide further information.

src/Controller/MainPageController.py
```diff
@@ -1184,6 +1184,13 @@ def __init__(self):
         self.dataset = self.patient_dict_container.dataset
         self.filepaths = self.patient_dict_container.filepaths
 
+    def reset(self):
```
tdlinh2712, line 1187: This works for this error, but this does not look like a good fix for me. This error shows that there are multiple instances of `PatientDictContainer` appear in the program at the same time, whereas this class is supposed to be Singleton, and this fix does not resolve that issue, and it only updates the patient_dict_container when anonymization is called.

I think the behavior that is supposed to happen is:
- There should be ONLY one PatientDictContainer instance inside the program at all times. Looks like all other classes are creating a shallow copy of the PatientDictContainer -> when the instance is updated, its shallow copies do not update. To resolve this, they should create deep copies, or keep the pointer to the singleton object (I'm more of a C++ user and not a Python user, so I'm not sure how to do it  in Python 😂, but that's the high-level idea)

- Another way is call reset() function for all classes that have a copy of PatientDictCotainer when the user clicks on open new patient to reset all PatientDictCotainer so that when opening a new patient, everything is reset. It's also weird to me that when opening a new patient, the program does not call _init_ functions of MainPageCallClass because it's supposed to re-setup all classes inside the MainPage?

tdlinh2712 (changes requested):

src/Controller/MainPageController.py
tdlinh2712: I see, yea I just searched about the difference in assigment between Python and C++ 😅
However, it means that self.patient_dict_container is updated, whereas `self.path`, `self.dataset`, and `self.filepaths` are not. So can we remove these variables to avoid inconsistency and use the value in the patient_dict_container instead?

for example:

`
    def run_anonymization(self, raw_dvh):
        target_path = anonymize(self.patient_dict_container.path, self.patient_dict_container.dataset, self.patient_dict_container.filepaths,
                                raw_dvh)
        return target_path
`

tdlinh2712: That way we don't need to call reset() function and all variables are always updated at all times for all functions in this class

Tamookk: While that would work, it would make the line length a lot longer than it needs to be, and would require that the other functions in `MainPageCallClass` be similarly updated. The class variables would also need to be removed as they would no longer be needed. This would require many more changes and testing to implement.

Having one reset function that can be called whenever a new patient is opened is a much cleaner way of implementing this fix, in my opinion.

tdlinh2712: @Tamookk I don't think so, because this leaves inconsistency in variables in the program. There should not be at anytime that there are two filepaths existing at the same time. This issue also happens in Clinical Data tab, because functions `def display_cd_dat` and `display_cd_form` use `tab_window, file_path, self.dataset, self.filepaths` as well, and if future developers add more functions to this class, they will have to be aware of this inconsistency.

To resolve the lengthy line, you can have local variables inside the function to improve readability:
`def run_anonymization(self, raw_dvh): `
`  path = self.patient_dict_container.path`
`  dataset = self.patient_dict_container.dataset`
`  filepaths = self.patient_dict_container.filepaths`
`  target_path = anonymize(path, dataset, filepaths, raw_dvh)`
`  return target_path`

Please consider this because it will prevent further bugs caused by unsynchronisation in the future.

Tamookk: Good thinking. I didn't consider that future functions would require that future developers be aware of this inconsistency. I have changed `MainPageCallClass` to remove the class variables and reset function, and have it simply grab the latest information from `PatientDictContainer` whenever it is required. These changes have been tested and the code appears to still work.

tdlinh2712: Awesome! Thanks!

tdlinh2712 (approved):

JThomas2020 (approved):
Team 19

PR #150, opened 2021-08-17
Add Isodose to ROI Functionality

Tamookk:
This pull request adds Isodose to ROI (ISO2ROI) functionality to OnkoDICOM. This allows users to generate and save ROIs into the RT Struct file of a dataset. These ROIs come from isodose levels calculated from data contained within RT Dose files.

An overview of what has been added and changed for this functionality:
- Isodose levels can be saved as ROIs.
- GUI has been added to the isodoses tab to allow users to start the ISO2ROI process.
- `isodoseRoi.csv` has been edited to include units.
- A screen has been added in the add-on options menu to allow users to create, modify, and delete isodose levels that they want included in the ISO2ROI process.
- An RT Struct file will be generated as a part of the ISO2ROI process if a dataset does not currently contain one.
- Unit tests for the ISO2ROI code have been created.

Some general changes also included in this pull request:
- The function to generate RT Struct files from a CT image has been slightly modified.
- The progress window class has been generalised so it can be used with more classes (in this instance, when performing the ISO2ROI conversion).
- Extra checks have been added to the main page before displaying the structures tab to make sure the required data exists.
- Various bug fixes.


sjswerdloff:
How many points per curve are typical? 
What precision is being used for each coordinate?
There is a weak point in DICOM with respect to VR of DS with VM of n (1-n,2-n,3-3n, etc) when using explicit endian syntax, such that the value of the the element can't exceed 64kbytes total because a byte length descriptor is 16 bits.  A naive conversion from Decimal to String can consume 16 bytes, x3 per point, and 1500 points later results in data corruption down the line. Even if we use implicit little endian, someone down the line might save with explicit or transmit with explicit syntax. 
Been there, done that, wrote the documents to help prevent the issue in a commercial environment (cough) 

It is not necessary to have precision more than 0.001 mm.  A blood cell is about. 0.008 mm wide, while muscle and parenchyma are about 0.030 mm.

If it is possible to vectorise to some extent, that's a good thing, but at least keep an eye on bytes per element. 
If converting to string in local code, be careful to use en-usa (dot vs comma) encoding.

data/csv/isodoseRoi.csv
```diff
@@ -1,11 +1,6 @@
-7310,ISO7310,
```
tdlinh2712, line 1: is this a desired modification to isodoseROI.csv?

src/Model/ISO2ROI.py
```diff
@@ -0,0 +1,292 @@
...
+
+                patient_dict_container.set("selected_rois", [])
+                patient_dict_container.set("dict_polygons_axial", {})
+
+        # Get isodose levels to turn into ROIs
+        self.get_iso_levels()
+
+        # Stop loading
+        if interrupt_flag.is_set():
+            print("Stopped ISO2ROI")
```
tdlinh2712, line 100: remove this print

src/Model/ISO2ROI.py
```diff
@@ -0,0 +1,292 @@
...
+            return False
+
+        # Calculate dose boundaries
+        progress_callback.emit(("Calculating Boundaries", 50))
+        boundaries = self.calculate_boundaries()
+
+        # Return if boundaries could not be calculated
+        if not boundaries:
+            print("Boundaries could not be calculated.")
```
tdlinh2712, line 109: Do we need these prints? in line 109 and 114

data/csv/isodoseRoi.csv
Tamookk: Units are now required in `isodoseRoi.csv`, as users can choose between creating levels in cGy and levels in %. Some isodose levels were removed during testing to speed up the ISO2ROI conversion process. These can be re-added if necessary.

Tamookk:
> How many points per curve are typical?

I have not seen any more than a thousand points per ROI (per slice) during development, and that's for the bigger ROIs. This number includes x, y, and z values.

> What precision is being used for each coordinate?

Coordinates are generated first by calculating contours from a dose grid. Points in these contours are then transformed using dose pixel look-up tables to convert them into pixel values. These pixel values are then converted into RCS coordinates. I would assume then that the precision of each coordinate is the same as the precision of the generated RCS. The mathematics behind this conversion is beyond me - I was just using what was already available within OnkoDICOM - so I am not sure what the exact precision of this would be. It may be to do with the pixel spacing specified in CT images?

src/Model/ISO2ROI.py
Tamookk: Print statements are included when the ISO2ROI process has to be prematurely stopped for any reason. They were included as an indication to the user of this, as we thought a simple print statement would be a lot cleaner code-wise than including extra pop-up windows, plus print statements do not interrupt the user and what they are doing. If multiple reviewers think that these are not necessary then they can be removed, however we do think that they should stay for reporting purposes.

tdlinh2712: Hm I see... I don't think print could replace a pop-up window if there's an error... Future developers might just think it's a debug print and delete it. Besides, when you packae it into - for example, a .exe program - users cant see the print if running this functionality has error.

Does this action happen in the background (while calculating boundaries user can still interact with other parts of the program)?

sjswerdloff: That's what logging is for.
You write to logs in well known directories (frequently by configuration).
Most trouble shooting in the commercial environment is by reading logs.  And if one can not determine the problem from the logs, one can at least determine where things stopped working.
Logging is your friend.  Logging with levels of severity is your life raft.

In general, the message to the user has to be relatively non technical.  Logs can be considered a note to yourself, although it's best for it to be understandable to fellow programmers, if not support personnel.  I dump stack traces in logs. I say naughty things about other vendors in my logs (when it's evidence that they can't count, for example).

Start using logging.

Tamookk: > Hm I see... I don't think print could replace a pop-up window if there's an error... Future developers might just think it's a debug print and delete it. Besides, when you package the program into - for example, a .exe file - users cant see the print if running this functionality has error.

That is true! However as @sjswerdloff says, logging would be better for errors. However adding logging is a big feature in and of itself, and we have other features to focus on that the product owner has requested. Perhaps this could be put to him?

For now the print statements can be deleted, or replaced with dialog boxes, if the majority think that is best.

JThomas2020 (commented):
Encountered an issue when deleting an ROI and subsequently converting any ISO to ROI:

`Traceback (most recent call last):
  File "/home/josh/team_14/OnkoDICOM/src/Controller/GUIController.py", line 93, in update_ui
    self.setup_central_widget()
  File "/home/josh/team_14/OnkoDICOM/src/View/mainpage/MainPage.py", line 98, in setup_central_widget
    self.structures_tab = StructureTab()
  File "/home/josh/team_14/OnkoDICOM/src/View/mainpage/StructureTab.py", line 25, in __init__
    self.color_dict = self.init_color_roi()
  File "/home/josh/team_14/OnkoDICOM/src/View/mainpage/StructureTab.py", line 78, in init_color_roi
    roi_id = self.patient_dict_container.get("list_roi_numbers")[int(id)]
IndexError: list index out of range
`

To replicate

1. Open OnkoDICOM and load a patient, click yes on loading rtstruct and rtdose
2. Go to Structures tab and click Delete ROI
3. follow the steps to delete at least one ROI
3. Go to Isodose tab and click Convert Isodose to ROI

Tamookk:
We are unable to replicate this with fresh installs on both Windows and Linux (Ubuntu and Mint).

JThomas2020 (commented):
Error from before is gone with a fresh clone

src/Model/ISO2ROI.py
didymo: there is now a sqlite file available for logs, is there not?? It is not a user, feature as was pointed out, messages to the screen are for the users, messages to logs are for the technical people. Make life easier for yourself and those who follow. DO not put in logging everywhere, only as the system grows, and put in log level flags from the beginning, with a defined log level in the config file so it is able to be altered.

https://docs.python.org/3/library/logging.html

https://docs.python.org/3/howto/logging.html

google returned several references of using sqlite with the python logger, and logging levels.

Tamookk: I did a quick look through the code base and found no mention of logging currently in OnkoDICOM - only in some of the libraries used. I've also been informed since that the print statements were originally as, during development, sometimes the ISO2ROI process would stop with no warning and no indication as to why. Since it all appears to be working now, these can be removed.

src/Model/ISO2ROI.py
```diff
@@ -0,0 +1,292 @@
...
+                patient_dict_container.set("pixluts", dict_pixluts)
+
+                rois = ImageLoading.get_roi_info(ds)
+                patient_dict_container.set("rois", rois)
+
+                patient_dict_container.set("selected_rois", [])
+                patient_dict_container.set("dict_polygons_axial", {})
+
+        # Get isodose levels to turn into ROIs
+        self.get_iso_levels()
```
tdlinh2712, line 96: a get function suggests it would return something, and `calculate_boundaries()` requires the latest iso_levels list, it's better to do something like:
```
iso_levels = self.get_iso_levels()
boundaries = self.calculate_boundaries(iso_levels)
```
and remove `self.isodose_levels` as well.

Let `get_iso_levels()` function return the `iso_levels` list instead of saving it to the class will help to avoid synchronization issues for the future, and put `iso_levels` as a parameter for `calculate_boundaries()` makes this function easier to reuse i think :D

src/Model/ISO2ROI.py
tdlinh2712: Looks like these print statements everywhere in the program so i think if we refactor it - better refactor it all in a future work item. If you keep these print. please add a comment in the code, like `# TODO: convert all print to logging`

src/Model/ISO2ROI.py
```diff
@@ -0,0 +1,292 @@
...
+        self.isodose_levels = {}
+
+        # Open isodoseRoi.csv
+        with open('data/csv/isodoseRoi.csv', "r") as fileInput:
+            for row in fileInput:
+                items = row.split(',')
+                self.isodose_levels[items[2]] = [items[1] == 'cGy',
+                                                 int(items[0])]
+
+    def calculate_boundaries(self):
```
tdlinh2712, line 140: maybe rename this to `calculate_roi_boundaries`?

src/Model/ISO2ROI.py
```diff
@@ -0,0 +1,292 @@
...
+
+        progress_callback.emit(("Validating Datasets", 0))
+        val = ImageLoading.is_dataset_dicom_rt(patient_dict_container.dataset)
+
+        # Stop loading
+        if interrupt_flag.is_set():
+            print("Stopped ISO2ROI")
+            return False
+
+        if not val:
```
tdlinh2712, line 47: what does `val` refer to?

src/Model/ISO2ROI.py
```diff
@@ -0,0 +1,292 @@
...
+                dict_pixluts = ImageLoading.get_pixluts(
+                    patient_dict_container.dataset)
+                patient_dict_container.set("pixluts", dict_pixluts)
+
+                rois = ImageLoading.get_roi_info(ds)
+                patient_dict_container.set("rois", rois)
+
+                patient_dict_container.set("selected_rois", [])
+                patient_dict_container.set("dict_polygons_axial", {})
+
```
tdlinh2712, line 94: Since these lines of code referring to creating a new RTSTRUCT file, it's better to put it in a separate function i.e `create_new_rtstruct()` in this class and calls the function here, for readabilitty

src/View/OpenPatientProgWindow.py
```diff
@@ -0,0 +1,82 @@
+import platform
```
tdlinh2712, line 1: OpenPatientProgressWindow? everywhere else uses progress though

src/Controller/AddOnOptionsController.py
```diff
@@ -261,33 +261,38 @@ def onCustomContextMenuRequestedVolume(self, pos):
...
-    #         dialog = Dialog_Dose(self.table_roi.item(
-    #             c, 0).text(), self.table_roi.item(c, 2).text())
-    #         if dialog.exec():
-    #             new_data = dialog.getInputs()
-    #             self.table_roi.setItem(c, 0, QTableWidgetItem(new_data[0]))
-    #             self.table_roi.setItem(c, 1, QTableWidgetItem(new_data[1]))
-    #             self.table_roi.setItem(c, 2, QTableWidgetItem(new_data[2]))
+    # ROI from IsoDoses
+    @Slot(QtCore.QPoint)
+    def onCustomContextMenuRequested_Roi(self, pos):
```
tdlinh2712, line 266: I'm a bit confused, is this supposed to be `snake_case` or `camelCase`? :joy: better modify it for consistency?

tdlinh2712 (commented):
Team 23:
Impressive work! I only have a few comments about naming convention and coding style.

There's one issue:
When I try to open a dataset that does not have a RTDOSE file (ie RT-02), or if I deselect RTDOSE for full RT set, I got this error:
```
<class 'AttributeError'> 'MainWindow' object has no attribute 'isodoses_tab' <traceback object at 0x7f2b58ef8dc0>
  File "/home/linh/OnkoDICOM/src/Controller/TopLevelController.py", line 76, in show_main_window
    self.main_window.update_ui()
  File "/home/linh/OnkoDICOM/src/Controller/GUIController.py", line 97, in update_ui
    self.isodoses_tab.request_update_ui.connect(self.update_ui)
AttributeError: 'MainWindow' object has no attribute 'isodoses_tab'
```

src/Model/ISO2ROI.py
Tamookk: That is a good idea, however I'm not sure if there would be another use-case in the future for reading the `isodoseRoi.csv` file outside of calculating isodose ROIs and populating the table in the options menu. Despite this, I will implement these changes.

src/Model/ISO2ROI.py
Tamookk: This function does not calculate ROI boundaries, it calculates isodose boundaries. However, it will be renamed to `calculate_isodose_boundaries`. As for the print statements, since enough people have commented on them they will be removed.

src/Model/ISO2ROI.py
Tamookk: `val` is a boolean; if it is True, it means that a loaded dataset/patient does not have at least one each of RT Dose, RT Plan, RT Struct, and CT Image. I understand that it is named poorly however and this will be fixed.

src/Model/ISO2ROI.py
Tamookk: That is a good idea. I will implement that.

src/View/OpenPatientProgWindow.py
Tamookk: This class inherits from the generic `ProgressWindow` class. This is because the `OpenPatientProgWindow` class has functions specific to the progress window that is created when opening a patient. Or so I would assume! I didn't work on this specific part of the code. The decision-making process behind this is only known to @LWMcClymont

Tamookk:
> There's one issue:
> When I try to open a dataset that does not have a RTDOSE file (ie RT-02), or if I deselect RTDOSE for full RT set, I got this error:
> 
> ```
> <class 'AttributeError'> 'MainWindow' object has no attribute 'isodoses_tab' <traceback object at 0x7f2b58ef8dc0>
> Traceback (most recent call last):
>   File "/home/linh/OnkoDICOM/src/Controller/TopLevelController.py", line 76, in show_main_window
>     self.main_window.update_ui()
>   File "/home/linh/OnkoDICOM/src/Controller/GUIController.py", line 97, in update_ui
>     self.isodoses_tab.request_update_ui.connect(self.update_ui)
> AttributeError: 'MainWindow' object has no attribute 'isodoses_tab'
> ```

A fix has been made for this and added to the PR.

src/View/OpenPatientProgWindow.py
tdlinh2712: Yea, I only suggest to rename this class to `OpenPatientProgressWindow` to improve readability

Tamookk: Ah! Okay, I get it. My bad. This change will also be made.

src/Model/ISO2ROI.py
didymo: > Looks like these print statements everywhere in the program so i think if we refactor it - better refactor it all in a future work item. If you keep these print. please add a comment in the code, like `# TODO: convert all print to logging`

An excellent way forward.

src/Model/ISO2ROI.py
Tamookk: Change has been made and added to the PR.

src/Model/ISO2ROI.py
Tamookk: Changes have been made and added to the PR.

src/Model/ISO2ROI.py
Tamookk: Change has been made and added to the PR.

src/Model/ISO2ROI.py
Tamookk: Change has been made and added to the PR.

src/View/OpenPatientProgWindow.py
Tamookk: Change has been made and added to the PR.

src/Controller/AddOnOptionsController.py
Tamookk: Change has been made and added to the PR.

src/Model/ISO2ROI.py
Tamookk: `TODO` comments have been added above print statements.

src/View/mainpage/IsodoseTab.py
```diff
@@ -17,12 +19,29 @@ def __init__(self):
         self.color_squares = self.init_color_squares()
         self.checkboxes = self.init_checkboxes()
 
+        # Create and initialise ISO2ROI button and layout
+        self.iso2roi_button = QtWidgets.QPushButton()
+        self.iso2roi_button.setText("Convert Isodose to ROI")
```
tdlinh2712, line 24: Since you convert many Isodoses to ROIs, should this be "Convert Isodoses to ROIs"? Same thing for the title in the Add-On Options

Tamookk: Changes have been committed.

JThomas2020:
Encountered an issue when converting iso to roi after deleting all other rois:

`ERROR:root:Traceback (most recent call last):
  File "/home/josh/team_14/OnkoDICOM/src/Model/Worker.py", line 77, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/josh/team_14/OnkoDICOM/src/Model/ISO2ROI.py", line 85, in start_conversion
    self.generate_roi(boundaries, progress_callback)
  File "/home/josh/team_14/OnkoDICOM/src/Model/ISO2ROI.py", line 231, in generate_roi
    rtss = ROI.create_roi(dataset_rtss, item,
  File "/home/josh/team_14/OnkoDICOM/src/Model/ROI.py", line 172, in create_roi
    rtss.ReferencedFrameOfReferenceSequence[0].FrameOfReferenceUID
  File "/home/josh/team_14/OnkoDICOM/venv/lib/python3.8/site-packages/pydicom/dataset.py", line 836, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'FileDataset' object has no attribute 'ReferencedFrameOfReferenceSequence'
  File "/home/josh/team_14/OnkoDICOM/src/Model/Worker.py", line 77, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/josh/team_14/OnkoDICOM/src/Model/ISO2ROI.py", line 85, in start_conversion
    self.generate_roi(boundaries, progress_callback)
  File "/home/josh/team_14/OnkoDICOM/src/Model/ISO2ROI.py", line 231, in generate_roi
    rtss = ROI.create_roi(dataset_rtss, item,
  File "/home/josh/team_14/OnkoDICOM/src/Model/ROI.py", line 172, in create_roi
    rtss.ReferencedFrameOfReferenceSequence[0].FrameOfReferenceUID
  File "/home/josh/team_14/OnkoDICOM/venv/lib/python3.8/site-packages/pydicom/dataset.py", line 836, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'FileDataset' object has no attribute 'ReferencedFrameOfReferenceSequence'`

To replicate (I'd recommend creating a copy of the test data to restore later):

1. Open up the patient in DICOM-RT-02
2. On Structures tab, go into the Delete ROI tab and delete all ROIs
3. Go to Isodose and click 'Convert Isodose to ROI.' Program will be stuck on 75% and error message occurs

Note this doesn't crash the program and all other functionality is still available, however Isodoses cannot be converted to ROI due to this error

Tamookk:
This was caused due to a recent change to `create_roi` in `ROI.py` and has been fixed.

JThomas2020 (approved):
Team 19, Good Job Guys!

src/Model/ImageLoading.py
```diff
@@ -514,3 +514,19 @@ def get_pixluts(read_data_dict):
...
+    :return: uid_list, a list of SOPInstanceUIDs of all image slices of the patient
+    """
+    non_img_list = ['rtss', 'rtdose', 'rtplan', 'rtimage']
+    uid_list = []
+
+    # Extract the SOPInstanceUID of every image (except RTSS, RTDOSE, RTPLAN)
+    for key in dataset:
+        if key not in non_img_list:
+            uid_list.append(dataset[key].SOPInstanceUID)
+    return uid_list
```
DuyTungHa, line 532: PEP-8 No new line at end of file.

src/Model/ROI.py
```diff
@@ -692,96 +706,89 @@ def create_initial_rtss_from_ct(img_ds: pydicom.dataset.Dataset,
...
+
+    # Add required elements
+    rt_ss.InstanceCreationDate = dicom_date
+    rt_ss.InstanceCreationTime = dicom_time
 
-    rt_ss.InstanceCreationTime = rt_ss.StructureSetTime = dicom_time
     rt_ss.is_little_endian = True
     rt_ss.is_implicit_VR = True
-    return rt_ss
+    return rt_ss
```
DuyTungHa, line 794: PEP-8 New line

src/Model/ImageLoading.py
Tamookk: Fixed

src/Model/ROI.py
Tamookk: Fixed

DuyTungHa (approved):
If the user does not select an existing RTSS file and choose to perform ISO2ROI then a new RTSS file will be created in the same directory. This violates the rule that only one RTSTRUCT file can be in a study. However, the merging of RTSS files is the work of Team 23 (part of the functionality "create ROI if no RTSTRUCT exists").

PR #155, opened 2021-08-28
ROI Create (where there is no RTSTRUCT existing)

DuyTungHa:
**Expected Functionality**
The user will select an image set normally (without RTSTRUCT).
IF the selected image set has a RTSTRUCT file associated with it, that link will be recognised.
The selected image set will open in the normal window. The Structure Tab on the left will be visible but empty.
The user will select the CreateROI and see the normal window, and be able to draw a named ROI as normal.
The user will save the created ROI, and:

- IF there is no RTSTRUCT, a new RTSTRUCT file will be created in the same directory and linked to the primary image set.
- IF there is a RTSTRUCT and the new ROI has a unique name, the ROI will be added to that RTSTRUCT.
- IF there is a RTSTRUCT and the new ROI is NOT unique name, the GUI will inform the user and ask the user to rename or permission to overwrite the existing ROI.

**Notes**
- This PR also fixes the memory leak issue in DrawROIWindow and DeleteROIWindow. Both are subclasses of QMainWindow. The program will now have only one instance for each of DrawROIWindow and DeleteROIWindow. The UI of these windows will be updated when a new patient is opened. (The previous behavior is that a new instance will be created). 
- The auto save functionality of ISO2ROI is kept the same as before except for the case where there is a RTSTRUCT existing and the new ROI is NOT unique name. In this case, the program will ask user for permission to overwrite the existing ROI.


JThomas2020 (approved):
Team 19

src/Model/InitialModel.py
```diff
@@ -90,25 +90,27 @@ def create_initial_model():
...
-            ImageLoading.get_raw_contour_data(dataset['rtss'])
-        patient_dict_container.set("raw_contour", dict_raw_contour_data)
-
+    patient_dict_container.set("file_rtss", filepaths['rtss'])
+    patient_dict_container.set("dataset_rtss", dataset['rtss'])
+    dict_raw_contour_data, dict_numpoints = \
+        ImageLoading.get_raw_contour_data(dataset['rtss'])
+    patient_dict_container.set("raw_contour", dict_raw_contour_data)
+
+    # dict_dicom_tree_rtss will be set in advance if the program generates a new rtss
```
Tamookk, line 99:
```suggestion
    # dict_dicom_tree_rtss will be set in advance if the program
    # generates a new rtss
```

PEP 8 compliance - comment line too long.

src/Model/ROI.py
```diff
@@ -792,3 +789,66 @@ def create_initial_rtss_from_ct(img_ds: pydicom.dataset.Dataset, filepath: Path,
     rt_ss.is_little_endian = True
     rt_ss.is_implicit_VR = True
     return rt_ss
+
+
+def merge_rtss(old_rtss, new_rtss, duplicated_names):
+    """
+    Merge two RTSTRUCT datasets.
+    Replace the ROIs in old_rtss with the ROIs in new_rtss which have same names.
```
Tamookk, line 797:
```suggestion
    Replace the ROIs in old_rtss with the ROIs in new_rtss which have
    same names.
```

PEP 8 compliance - comment line too long.

src/Model/ROI.py
```diff
@@ -792,3 +789,66 @@ def create_initial_rtss_from_ct(img_ds: pydicom.dataset.Dataset, filepath: Path,
...
+
+    # Merge the original sequences with the new sequences
+    original_structure_set.extend(new_structure_set)
+    original_roi_contour.extend(new_roi_contour)
+    original_roi_observation_sequence.extend(new_roi_observation_sequence)
+
+    # Renumber the ROINumber and ReferencedROINumber tags
+    original_structure_set = renumber_roi_number(original_structure_set)
+    original_roi_contour = renumber_roi_number(original_roi_contour)
+    original_roi_observation_sequence = renumber_roi_number(original_roi_observation_sequence)
```
Tamookk, line 836:
```suggestion
    original_roi_observation_sequence = \
        renumber_roi_number(original_roi_observation_sequence)
```

PEP 8 compliance - line too long (also lines 839, 841).

src/View/ImageLoader.py
```diff
@@ -16,10 +19,11 @@ class ImageLoader(QtCore.QObject):
 
     signal_request_calc_dvh = QtCore.Signal()
 
-    def __init__(self, selected_files, parent_window, *args, **kwargs):
+    def __init__(self, selected_files, existing_rtss_path, parent_window, *args, **kwargs):
```
Tamookk, line 22:
```suggestion
    def __init__(self, selected_files, existing_rtss_path, parent_window, 
                 *args, **kwargs):
```

PEP 8 compliance - line too long (also lines 46-47, 134-137, 142-143).

src/View/OpenPatientWindow.py
```diff
@@ -324,9 +325,24 @@ def tree_item_clicked(self, item, _):
             header = "Only one study can be opened."
             self.open_patient_window_confirm_button.setDisabled(True)
 
+        # If RTSTRUCT exists but not selected, save existing RTSTRUCT file path
+        self.existing_rtss_path = self.get_existing_rtss_path(selected_series_types)
```
Tamookk, line 329:
```suggestion
        self.existing_rtss_path = \
            self.get_existing_rtss_path(selected_series_types)
```

PEP 8 - line too long (also lines 339, 342, 358).

src/View/mainpage/IsodoseTab.py
```diff
@@ -166,5 +166,5 @@ def on_loaded_iso2roi(self):
         Closes the progress window and refreshes
         the main screen.
         """
-        self.request_update_ui.emit()
+        self.request_update_ui.emit((self.patient_dict_container.get('dataset_rtss'), {"draw": None}))
```
Tamookk, line 169:
```suggestion
        self.request_update_ui.emit((
            self.patient_dict_container.get('dataset_rtss'), {"draw": None}))
```

PEP 8 compliance - line too long.

src/View/mainpage/MainPage.py
```diff
@@ -89,37 +86,21 @@ def setup_central_widget(self):
...
-                 and patient_dict_container.get("rtss_modified") is True):
-                self.structures_tab = StructureTab()
-                self.structures_tab.request_update_structures.connect(
-                    self.update_views)
-                self.left_panel.addTab(self.structures_tab, "Structures")
-        elif hasattr(self, 'structures_tab'):
-            del self.structures_tab
+        if not hasattr(self, 'structures_tab'):
+            self.structures_tab = StructureTab()
+            self.structures_tab.request_update_structures.connect(self.update_views)
```
Tamookk, line 91:
```suggestion
            self.structures_tab.request_update_structures.connect(
                self.update_views)
```

PEP 8 compliance - line too long (also line 99).

src/View/mainpage/StructureTab.py
```diff
@@ -1,15 +1,16 @@
...
-
 from PySide6 import QtWidgets, QtGui, QtCore
 from PySide6.QtCore import Qt
 
 from src.Controller.ROIOptionsController import ROIDelOption, ROIDrawOption
 from src.Model import ImageLoading
 from src.Model.GetPatientInfo import DicomTree
 from src.Model.PatientDictContainer import PatientDictContainer
-from src.Model.ROI import ordered_list_rois, get_roi_contour_pixel, calc_roi_polygon, transform_rois_contours
+from src.Model.ROI import ordered_list_rois, get_roi_contour_pixel, calc_roi_polygon, transform_rois_contours, \
```
Tamookk, line 12: PEP 8 compliance - line too long (also lines 357, 361, 369-373, 377, 381-382, 386, 393, 401, 403, and 407-409).

test/test_model_roi.py
```diff
@@ -4,10 +4,63 @@
...
+                dcmread(os.path.join(root, name))
+            except (InvalidDicomError, FileNotFoundError):
+                pass
+            else:
+                dicom_files.append(os.path.join(root, name))
+    return dicom_files
+
+
+class TestROI:
+    """Class to set up the OnkoDICOM main window for testing the structures tab."""
```
Tamookk, line 42: Line too long - PEP 8 compliance (also lines 48-50, 55, and 156-157).

test/test_view_structures_tab.py
```diff
@@ -1,11 +1,15 @@
...
+
+import pydicom
 import pytest
 from pathlib import Path
+from PySide6 import QtWidgets, QtCore
 
 from src.Controller.GUIController import MainWindow
 from src.Model.PatientDictContainer import PatientDictContainer
-from src.Model.ROI import get_contour_pixel, calc_roi_polygon
+from src.Model.ROI import get_contour_pixel, calc_roi_polygon, create_initial_rtss_from_ct, create_roi, merge_rtss
```
Tamookk, line 10: Line too long - PEP 8 compliance (also lines 59, 127-128, 160-163, 170, 177-178, 190, and 209).

Tamookk (commented):
Mostly works well! My main complaints are that some of the line lengths are too long for PEP 8 compliance.

Only encountered one problem:

```
<class 'FileNotFoundError'> [Errno 2] No such file or directory: 'D:\\Downloads\\DICOM Test Data\\DICOM-RT-02\\rtss.dcm' <traceback object at 0x000002240AA71B80>
  File "C:\Users\Joshua\Desktop\OnkoDICOM\src\View\mainpage\StructureTab.py", line 266, in structure_modified
    self.save_new_rtss(auto=True)
  File "C:\Users\Joshua\Desktop\OnkoDICOM\src\View\mainpage\StructureTab.py", line 380, in save_new_rtss
    old_rtss = pydicom.dcmread(existing_rtss_directory, force=True)
  File "C:\Users\Joshua\Desktop\OnkoDICOM\venv\lib\site-packages\pydicom\filereader.py", line 1001, in dcmread
    fp = open(fp, 'rb')
FileNotFoundError: [Errno 2] No such file or directory: 'D:\\Downloads\\DICOM Test Data\\DICOM-RT-02\\rtss.dcm'
```

Steps to reproduce:

1. On the open patient window, select a folder that contains an RT Struct. Wait until searching has finished, and the files in the directory are shown (including the RT Struct).
2. Delete the RT Struct from the directory.
3. Select all files, except the RT Struct.
4. Continue loading the patient.


While I am not sure how often this problem would pop up in the real world (it would require a directory to have been fully read, and then the RTSS deleted and not selected when opening the patient), it does raise some interesting questions.

It appears as though the RT Struct is being opened even if it is not selected in the Open Patient window. As far as I understand, if the original RT Struct isn't selected, then a new one is created and the data from the old one is added to it. If this is the case - why not just open the original RT Struct and add data straight to it?

If it's always going to be opened whether it is selected or not, then why give the user a choice? Why create a blank structures tab instead of just opening the RT Struct, and using the data contained in it, if you're going to open it and use it to make a new RT Struct anyway?

I'm quite impressed by the work done! I'm just confused as to the reasoning behind how and why these changes were implemented.

src/Model/InitialModel.py
minhhuyxx: fixed

minhhuyxx:
@Tamookk. Thanks for your comments.
I have fixed the code to comply with PEP8.

DuyTungHa:
@Tamookk 
The problem you encountered there is because we cache the existing rtss file path if user does not select that file so that we can merge the ROIs latter on. Because the rtss was deleted after the searching has finished, the cached path now does not exist anymore. This happens because onkoDICOM only does searching once and does not reflect changes made after searching is done. 

I can implement a safer merge ROIs function where if the path is no longer valid, no merging will be done and the new rtss is automatically saved. But I'm not sure if it is worth it because I think it is very unlikely that this problem will ever occur in real world. 

In response to your second question, there is an use case where the user just want to create new ROIs and do not wish to look at the existing ROIs (probably because of performance). That's why we create a temporary rtss in memory and when user hit save, it will check whether there is an existing rtss. If there is, merging will be performed between two RTSTRUCTs, otherwise, the temporary rtss will be saved as a new rtss in the file system. 

Hope this answers your questions.

Tamookk (approved):
Thank you for the explanation, that all makes sense to me. Good work!

PR #157, opened 2021-09-02
3d vtk view

tdlinh2712:
This PR provide the 4th view in four-window view: 3D model.

Please note that this 3D view does not use the same zoom in-zoom out as other three views. to zoom in, you need to right click in the top right of the 3d window and to zoom out, right click in the bottom right of the 3d window.
Another way to zoom-in/out is to use the wheel of your mouse. 
This model can also render under different window levels (bone, soft tissues, lung etc).

Since this PR makes a change in requirements.txt, please do `pip install -r requirements.txt` before running the program.


src/View/mainpage/DicomView3D.py
```diff
@@ -0,0 +1,226 @@
+import numpy as np
+import vtk
+from PySide6 import QtWidgets
+from PySide6.QtWidgets import QPushButton
+from vtkmodules.util import numpy_support
+from vtkmodules.util.vtkConstants import VTK_INT
+from vtkmodules.vtkCommonDataModel import vtkImageData, vtkPiecewiseFunction
+from vtkmodules.vtkRenderingCore import vtkColorTransferFunction, vtkRenderer, vtkVolumeProperty, \
```
Tamookk, line 8: Too long for PEP 8 compliance, as are lines:
18, 47, 59, 62, 66, and 161.

src/View/mainpage/DicomView3D.py
```diff
@@ -0,0 +1,226 @@
...
+        self.iren.Initialize()
+        self.iren.Start()
+        self.vtk_widget.focusWidget()
+
+        self.is_rendered = True
+
+    def closeEvent(self, QCloseEvent):
+        """
+        Clean up function when the window is closed
+        """
```
Tamookk, line 211:
```suggestion
        :param QCloseEvent: ...
        """
```

src/View/util/QVTKRenderWindowInteractor.py
```diff
@@ -0,0 +1,491 @@
+import vtkmodules.qt
```
Tamookk, line 1: Just curious about this file. It seems to have been copied from VTK's source, right? Why not just include this file instead of copying it, were some modifications needed so it would work?

Tamookk (commented):
Good work! Just a couple PEP8 compliance problems, and one question regarding one of the files in this PR. 

I encountered no issues when testing. Very impressive!

src/View/util/QVTKRenderWindowInteractor.py
Regnum771: When we first started using VTK, the file required some modification to works with PySide6. It seems that VTK has been updated to work with PySide6 now. Thank you for bringing it up.

Tamookk: I just had a look myself and saw that VTK had been updated, but it seems as though it hasn't been updated on PyPi (so `pip install vtk` won't give you the newest version with PySide6 support). I think for now it's best to leave this in here, however since it's been copied from vtk and modified you will need to include the licence - just as a comment at the top of this file should be enough.

The licence file is here: https://gitlab.kitware.com/vtk/vtk/-/blob/master/Copyright.txt

eddieGong2604: @Tamookk @Regnum771 pip hasn't released the new version of vtk which includes Pyside6 support yet. I think we should stick to this workaround for now.

Regnum771: We will test the new version to see if it actually works. We'll include the copyright on top of the file as well. Thank you.

Tamookk: Leave what's there for now, just add the copyright.

What is currently in this PR will need to stay until pip has updated with the new version of VTK. We don't know when this will be, and what is in this PR works perfectly, so there's no need to test the new version for this PR.

Tamookk: For `DicomView3D.py`, lines 8, 66, 67, and 71 are still a little too long, and you still need to add `:param QCloseEvent: ...` in the docstring under the `closeEvent` function. Good work adding the licence and with the other lines though!

eddieGong2604: Can you please have a review again?

Tamookk: Perfect! Great work

Tamookk (approved):
Team 14 - great work! Very impressive

JThomas2020 (approved):
Team 19, Great Job Guys, I'll say is we should update the minimum requirements to run the 3D view, in Linux it uses approx 2GB to run the 3D view and up to 3.3GB for windowing.

PR #158, opened 2021-09-03
Modern-PT support

JThomas2020:
This PR adds support for modern-PT files into OnkoDICOM, specifically:

- Added a check for patient position data, which is optional in modern PT but mandatory in other DICOM image types
- Added a byte length parameter to address shear issues in generating an image slice (Big thanks to Team 14 for help with this)
- Added a function which transforms the transect input from 512x512 to the size of the pixel array

Please test with modern-PT and non PT to check robustness


DuyTungHa (approved):
Team23

Tamookk (approved):
Works well! Encountered no problems.

- Team 14

PR #160, opened 2021-09-05
Pt view Create and View ROI functionality for Modern CT

JThomas2020:
This PR adds the ability to create and View ROIs for modern PT and adds robustness to creating and viewing ROIs.

This comes from the investigation of a bug where modern CT ROI generation created index out of range errors, investigating revealed it wasn't the PT images themselves but their pixel size. They were not 512x512 but other square sizes (e.g. 256x256, 169x169). This PR adresses this issue by adding and using two functions to scale input positional data and ROI output. This will affect all ROI generation and viewing so please test different image types and ROIs

### DIFF
--- a/src/Model/ROI.py
+++ b/src/Model/ROI.py
@@ -12,6 +12,9 @@ from pydicom.uid import generate_uid, ImplicitVRLittleEndian
 from src.Model.CalculateImages import *
 from src.Model.PatientDictContainer import PatientDictContainer
 
+from src.View.mainpage.DrawROIWindow.Drawing import linear_transform, \
+    inv_linear_transform
+
 
 def rename_roi(rtss, roi_id, new_name):
     """
@@ -639,12 +642,21 @@ def calc_roi_polygon(curr_roi, curr_slice, dict_rois_contours, pixmap_aspect=1):
 
     list_polygons = []
     pixel_list = dict_rois_contours[curr_roi][curr_slice]
+    patient_dict_container = PatientDictContainer()
+    dataset = patient_dict_container.dataset[0]
+
+
     for i in range(len(pixel_list)):
         list_qpoints = []
         contour = pixel_list[i]
         for point in contour:
-            curr_qpoint = QtCore.QPoint(point[0], point[1] * pixmap_aspect)
-            list_qpoints.append(curr_qpoint)
+            x_t, y_t = inv_linear_transform(
+                point[0], point[1],
+                dataset['Rows'].value, dataset['Columns'].value)
+            for x in x_t:
+                for y in y_t:
+                    curr_qpoint = QtCore.QPoint(x, y * pixmap_aspect)
+                    list_qpoints.append(curr_qpoint)
         curr_polygon = QtGui.QPolygonF(list_qpoints)
         list_polygons.append(curr_polygon)
     return list_polygons
--- a/src/View/mainpage/DrawROIWindow/Drawing.py
+++ b/src/View/mainpage/DrawROIWindow/Drawing.py
@@ -66,11 +66,13 @@ class Drawing(QtWidgets.QGraphicsScene):
             if hasattr(self.draw_roi_window_instance, 'bounds_box_draw'):
                 bound_box = \
                     self.draw_roi_window_instance.bounds_box_draw.box.rect()
-                self.min_x, self.min_y = self.linear_transform(
-                    bound_box.x(), bound_box.y()
+                self.min_x, self.min_y = linear_transform(
+                    bound_box.x(), bound_box.y(),
+                    len(self.data), len(self.data[0])
                 )
-                self.max_x, self.max_y = self.linear_transform(
-                    bound_box.width(), bound_box.height()
+                self.max_x, self.max_y = linear_transform(
+                    bound_box.width(), bound_box.height(),
+                    len(self.data), len(self.data[0])
                 )
                 self.max_x += self.min_x
                 self.max_y += self.min_y
@@ -98,15 +100,16 @@ class Drawing(QtWidgets.QGraphicsScene):
             it. """
             # Convert QPixMap into Qimage
             for x_coord, y_coord in self.target_pixel_coords:
-                x_colour, y_colour = self.inv_linear_transform(
-                    x_coord, y_coord)
+                x_colour, y_colour = inv_linear_transform(
+                    x_coord, y_coord, len(self.data), len(self.data[0]))
                 c = self.q_image.pixel(x_colour[0], y_colour[0])
                 colors = QColor(c).getRgbF()
                 self.according_color_dict[(x_coord, y_coord)] = colors
             color = QtGui.QColor()
             color.setRgb(90, 250, 175, 200)
             for x_coord, y_coord in self.according_color_dict:
-                x_arr, y_arr = self.inv_linear_transform(x_coord, y_coord)
+                x_arr, y_arr = inv_linear_transform(
+                    x_coord, y_coord, len(self.data), len(self.data[0]))
                 for x in x_arr:
                     for y in y_arr:
                         self.q_image.setPixelColor(x, y, color)
@@ -139,24 +142,9 @@ class Drawing(QtWidgets.QGraphicsScene):
         """
         for i in range(constant.DEFAULT_WINDOW_SIZE):
             for j in range(constant.DEFAULT_WINDOW_SIZE):
-                x, y = self.linear_transform(i, j)
+                x, y = linear_transform(
+                    i, j, len(self.data), len(self.data[0]))
                 self.values.append(self.data[x][y])
-
-    # Scales DICOM view size to image size
-    def linear_transform(self, x, y):
-        m_x = float(len(self.data)) / constant.DEFAULT_WINDOW_SIZE
-        m_y = float(len(self.data[0])) / constant.DEFAULT_WINDOW_SIZE
-        return int(m_x * x), int(m_y * y)
-
-    # Sclaes image size to DICOM view size
-    def inv_linear_transform(self, x, y):
-        m_x = float(constant.DEFAULT_WINDOW_SIZE) / len(self.data)
-        m_y = float(constant.DEFAULT_WINDOW_SIZE) / len(self.data[0])
-        x_min = int(m_x * x)
-        x_max = int(m_x * (x + 1))
-        y_min = int(m_y * y)
-        y_max = int(m_y * (y + 1))
-        return numpy.arange(x_min, x_max), numpy.arange(y_min, y_max)
 
     def refresh_image(self):
         """
@@ -180,7 +168,8 @@ class Drawing(QtWidgets.QGraphicsScene):
         # The roi drawn on current slice is changed after several pixels are
         # modified
         self.slice_changed = True
-        clicked_x, clicked_y = self.linear_transform(clicked_x, clicked_y)
+        clicked_x, clicked_y = linear_transform(
+            clicked_x, clicked_y, len(self.data), len(self.data[0]))
         according_color_dict_key_list = list(self.according_color_dict.keys())
 
         color_to_draw = QtGui.QColor()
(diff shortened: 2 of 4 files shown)


DuyTungHa (commented):
Is there a way to check if an image is 512x512 or not ? Doing linear transformation is very time consuming and it significantly increases the amount of time required to render a ROI (especially for ROI with lots of contour data). It would be better if we only run linear transformation on images that do not meet this standard.

JThomas2020:
> Is there a way to check if an image is 512x512 or not ? Doing linear transformation is very time consuming and it significantly increases the amount of time required to render a ROI (especially for ROI with lots of contour data). It would be better if we only run linear transformation on images that do not meet this standard.

Yes, when an image is loaded, its dimensions are stored in patient_dict_container.dataset[x]["Rows" | "Columns"].value. OnkoDICOM automatically inflates the image to 512x512 in all viewers however this the only place its inflated, breaking fuctions which rely on pixel values and cooridinates as they're not in the same ranges.

The functions introduced in this PR translate the 512x512 input through the dicom views into NxN, where N < 512, which is fast. When pixels need to be updated on the views (and only the views), the functions generate every pixel that will be effected by the change and changes them into 512x512 output.

Another reason I implemented it this way is because I wanted to store/export the ROIs in the same dimensions as the original image and didn't want to manipulate the original image just to work with OnkoDICOM.

Also, ROIs should be quicker to generate as there are less coordinates to store for an NxN image than a 512x512 image (assuming N<512), and ROI rendering should take the same ammount of time (for x coordinates in an NxN image and sum(all coordinates x) <= N^2, inflating gives us x*((512/N)^2) coordinates so sum(all coordinates x*((512/N)^2)) <= (N^2)*((512/N)^2) = 512^2). Keep in mind most images are in the 512x512 format already so performing the transformation is the same as multiplying by 1

DuyTungHa:
Multiplying by 1 in terms of performance is not the same as not multiplying at all. What I mean is that we should do transformation only for cases where it is necessary to do so. Because the cost of just looping through every point in contour data and doing unnecessary transformation (x1) is huge. 
If you use DICOM-RT02 as a test dataset, you can visually feel the difference between displaying a ROI with no transformation and with transformation.

JThomas2020:
Good point, and yeah i see what you mean, i'll write a check now

src/Model/ROI.py
```diff
@@ -639,12 +641,23 @@ def calc_roi_polygon(curr_roi, curr_slice, dict_rois_contours, pixmap_aspect=1):
...
     pixel_list = dict_rois_contours[curr_roi][curr_slice]
+    dataset = PatientDictContainer().dataset[0]
+
     for i in range(len(pixel_list)):
         list_qpoints = []
         contour = pixel_list[i]
         for point in contour:
-            curr_qpoint = QtCore.QPoint(point[0], point[1] * pixmap_aspect)
-            list_qpoints.append(curr_qpoint)
+            if dataset['Rows'].value != DEFAULT_WINDOW_SIZE:
```
DuyTungHa, line 650: I think it would be even better if you can move this if else statement outside of the for loop because the dataset['Rows'].value is the same for all point in contour. So it is a bit redundant and time-consuming if the program have to check this for every point in the contour.

JThomas2020: Made a new variable to store the Boolean result, checks the variable rather than dataset['rows']

DuyTungHa (approved):
Team 23. 
Just one last comment: You might consider renaming variable or adding comments next to variables so that it is clear to future developer what the purpose of that variable is.

src/View/mainpage/DrawROIWindow/Drawing.py
```diff
@@ -303,14 +407,15 @@ def mousePressEvent(self, event):
...
+        x, y = linear_transform(
+            math.floor(event.scenePos().x()), math.floor(event.scenePos().y()),
+            self.rows, self.cols)
+        is_coloured = (x, y) in self.according_color_dict
         self.is_current_pixel_coloured = is_coloured
         self.draw_cursor(event.scenePos().x(), event.scenePos().y(),
                          self.draw_tool_radius, new_circle=True)
 
-        if self.is_current_pixel_coloured:
+        if not self.is_current_pixel_coloured:
```
tdlinh2712, line 418: ? Isn't this reversing the logic of the cursor in draw roi?? I don't think it's relevant to this PR

src/View/mainpage/DrawROIWindow/Drawing.py
```diff
@@ -329,7 +434,7 @@ def mouseMoveEvent(self, event):
         if self.cursor and self.isPressed:
             self.draw_cursor(event.scenePos().x(), event.scenePos().y(),
                              self.draw_tool_radius)
-            if self.is_current_pixel_coloured:
+            if not self.is_current_pixel_coloured:
```
tdlinh2712, line 437: here too

tdlinh2712 (changes requested):

src/View/mainpage/DrawROIWindow/Drawing.py
```diff
@@ -96,14 +103,31 @@ def _display_pixel_color(self):
             creates a new layer over the existing image instead of replacing 
             it. """
             # Convert QPixMap into Qimage
-            for x_coord, y_coord in self.target_pixel_coords:
-                c = self.q_image.pixel(x_coord, y_coord)
-                colors = QColor(c).getRgbF()
-                self.according_color_dict[(x_coord, y_coord)] = colors
+            if self.different_sizes:
+                for x_coord, y_coord in self.target_pixel_coords:
```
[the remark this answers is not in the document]
JThomas2020: I think it would be even better if you can move this if else statement outside of the for loop because the dataset['Rows'].value is the same for all point in contour. So it is a bit redundant and time-consuming if the program have to check this for every point in the contour.

_Originally posted by @DuyTungHa in https://github.com/didymo/OnkoDICOM/pull/160#discussion_r702541600_

I originally had the if statement in the code like that, but as @DuyTungHa pointed out, as the condition doesnt change for each point, keeping the condition outside speeds up the process

src/View/mainpage/DrawROIWindow/Drawing.py
```diff
@@ -277,9 +371,19 @@ def draw_contour_preview(self, list_of_points):
...
+                    point[0], point[1], self.rows, self.cols)
+                for x in x_arr:
+                    for y in y_arr:
+                        qpoint = QtCore.QPoint(x, y)
+                        qpoint_list.append(qpoint)
+        else:
+            for point in list_of_points:
+                qpoint = QtCore.QPoint(point[0], point[1])
+                qpoint_list.append(qpoint)
+
```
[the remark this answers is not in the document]
JThomas2020: Hi Linh, its a good thought but not what I intended, inv_linear_transform's purpose is to turn a 1*1 pixel into an n*n pixel to interact with the QImage, so while its possible, I don't think it will change the complexity as i need all n*n pixels generated.

src/View/mainpage/DrawROIWindow/Drawing.py
JThomas2020: Ah sorry, I switched it around when investigating and testing to add/remove points easier, i'll change it

src/View/mainpage/DrawROIWindow/Drawing.py
JThomas2020: same as above

src/View/mainpage/DrawROIWindow/Drawing.py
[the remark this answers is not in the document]
tdlinh2712: I see that now... However the trade off for that is this code repetition which is not ideal, especially when the repetition happen across multiple places in the code like this. It makes this file a lot more difficult to maintain... Can you think of anyway to resolve this?

Maybe when opening the drawing window, we transform the x and y coordinates first so that we dont have to re-transform them everytime there's a change in the drawing?

Then when we save the final coordinates, we can transform them back to the correct coordinates for saving?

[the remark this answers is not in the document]
tdlinh2712: Because if you transform the coordinates in this function, it will transform all drawn coordinates again and again everytime there's a change in the drawing, thus probably will slow down the drawing process for PET scan

src/View/mainpage/DrawROIWindow/Drawing.py
```diff
@@ -346,3 +451,20 @@ def mouseReleaseEvent(self, event):
...
         super().mouseReleaseEvent(event)
         self.update()
+
+
+# These functions scale window input to Image Size
+def linear_transform(x, y, m1, m2):
+    m_x = float(m1) / constant.DEFAULT_WINDOW_SIZE
+    m_y = float(m2) / constant.DEFAULT_WINDOW_SIZE
+    return int(m_x * x), int(m_y * y)
+
```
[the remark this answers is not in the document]
JThomas2020: Good idea, removing reused function and collecting them into src/Model/Transform.py

src/View/mainpage/DrawROIWindow/Drawing.py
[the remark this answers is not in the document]
JThomas2020: Currently most of the code hadles and saves the data as n\*n, while 512\*512 is used just to interact with the QImage. Although inv_linear_transform is O(n\*n), linear_transform is O(1). Most of the code to read and check conditions uses linear_transform and is O(1) and inv_linear_transform is only used to write to QImage. I will implement the transform_coord_to_point as that will reduce duplicate code, but thats the best optomisation I can do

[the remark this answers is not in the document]
tdlinh2712: Thanks!
Yes, tbh I'm more concerned about the duplication since to actually increase the performance we need to refactor it, and it's not in the scope of this PR. Hence I just want the code to be more maintainable so that in the future, it will be less of a  burden for the developer who refactors this function :)

src/Model/Transform.py
```diff
@@ -0,0 +1,44 @@
+import numpy
+
+from src.constants import DEFAULT_WINDOW_SIZE
+
+
+# A safe way of dividing numbers
+def safe_division_transform(x, y, n_x, n_y, d_x, d_y):
```
Tamookk, line 7: Docstrings missing in this file.

Tamookk (commented):
Works well! No complaints other than performance seems to have taken a little hit for the bigger ROIs on images that aren't 512x512, but that was expected. Also, docstrings are missing from `Transform.py`.

The styling for the checkboxes in the Select Patient Window seems to have disappeared, but that might just be me.

This solves a problem we were having with SUV2ROI generation as well, where the generated ROIs were not displaying properly in the DICOM View. We are very grateful for that!

src/Model/Transform.py
JThomas2020: Just saw this as I committed docstrings haha, thanks for picking up on it!

JThomas2020:
> Works well! No complaints other than performance seems to have taken a little hit for the bigger ROIs on images that aren't 512x512, but that was expected. Also, docstrings are missing from Transform.py.

Before team 23's comments the code was suprisingly slow, and for imahes thata arent 512 by 512, more pixels are added to the contour data, which is why it's slower

> The styling for the checkboxes in the Select Patient Window seems to have disappeared, but that might just be me.

Thats a bit odd, as none of these changes interact with the select patient window, I'm on linux and the styling hasn't changed

> This solves a problem we were having with SUV2ROI generation as well, where the generated ROIs were not displaying properly in the DICOM View. We are very grateful for that!

Thanks! Same thing happened to me when I initially updated the drawROI function, the pixels were displaying in the top left quadrant of the screen haha

Tamookk (approved):
> Thats a bit odd, as none of these changes interact with the select patient window, I'm on linux and the styling hasn't changed

Probably a problem on my end, then. I've been having a few of those lately...

Good work here! Team 14

tdlinh2712 (approved):
Looks good now! Had no issue when testing

PR #161, opened 2021-09-08
fix crash when using mouse wheel on 3D view

eddieGong2604:
This PR resolves bug #1895 on onko redmine

- Fix crash when user interacts with 3D view using mouse wheel


JThomas2020 (approved):
Team 19, nice and quick

Tamookk (approved):
Team 14

PR #163, opened 2021-09-12
Fix - Image Fusion 

peteqian:
This PR is to address the bugs [1897 ](https://redmine.didymodesigns.com.au/issues/1897), [1898](https://redmine.didymodesigns.com.au/issues/1898) and [1903](https://redmine.didymodesigns.com.au/issues/1903) related to Image Fusion (except for ROI and Windowing - this will still need to be implemented).

A summary of the fixes:
- Zoom in/out working
- Crosshairs work
- Message box has been enlarged
- Able to delete window
- Opening a new patient after Image Fusion from a previous patient, does not affect the selection of the moving image of new patient

Work to be done still
- Windowing to be applied on Image Fusion
- ROIs not being displayed

PR #164, opened 2021-09-13
Bug: Image Fusion Windowing

JThomas2020:
This PR adds windowing support to Image Fusion

This PR adresses Dr Miller's request to get windowing working for Image Fusion

This PR was built based of changes from PR #163 so this PR should be merged after that PR has merged


Tamookk:
Please merge with the master branch of the main repo. A lot of changes in this PR are shared with the recently pulled in changes, and I can't tell what changes are specific to this PR (at least not without comparing this branch with master).

When changing the window after Image Fusion while focused on the Image Fusion tab, OnkoDICOM stalled for about 30 seconds (which included receiving a "This program is not responding" message from Windows). The functionality works well, although it does run a little slowly. I don't know if this is just me though - we all know I've had problems on my end before...

If it takes this long for everybody then might I suggest adding a progress bar, or at least a pop-up window asking the user to please wait? This would alleviate any concerns a user may have after receiving a "program not responding" message.

JThomas2020:
> When changing the window after Image Fusion while focused on the Image Fusion tab, OnkoDICOM stalled for about 30 seconds (which included receiving a "This program is not responding" message from Windows).

Whats happening is the colour maps need to be re-generated when windowing is applied. this does take a long time for me as well because effectively two image sets (old and fused) are being compared to make a completely new third image set for the color overlay.

> If it takes this long for everybody then might I suggest adding a progress bar, or at least a pop-up window asking the user to please wait? This would alleviate any concerns a user may have after receiving a "program not responding" message.

I've made a confirmation dialog, allowing the user to opt out of fusion windowing and just apply regular windowing. The windowing process is slow in general so it would be worthwhile to add. This being said we should have a week for bug fixing in the next couple of weeks, so I'll add a progress window then

src/Model/MovingModel.py
```diff
@@ -134,10 +134,13 @@ def create_moving_model():
                                   dicom_tree_rtplan.dict)
 
 
-def read_images_for_fusion():
+def read_images_for_fusion(level=0, window=0):
```
Tamookk, line 137: Update docstrings.

Tamookk (commented):
Thank you for your explanation, makes sense to me. And thanks for updating the PR. All looks good and works great, just one function needs updated docstrings before I will approve.

src/Model/MovingModel.py
JThomas2020: Updated, cheers for the pickup!

Tamookk (approved):
Team 14

DuyTungHa (approved):
Team23

PR #166, opened 2021-09-17
Add Clinical Data to DICOM SR Functionality

Tamookk:
This PR adds Clinical Data to DICOM SR functionality. An overview of the changes made:

- The existing clinical data form and exporting have been removed.
- The first time welcome window has been changed to allow the user to select a CSV file to gather patient clinical data from. The path of this CSV file is written to the same database that the default directory is written to (this can be skipped).
- An option in the add-on options menu has been added that allows users to change the file path of the CSV file to load patient clinical data from.
- When a dataset is opened, its ID is checked against the list of IDs in the patient clinical data CSV. If these IDs match, the table in the clinical data tab is automatically populated with the matching data in the CSV file.
- When clinical data is imported from a CSV, a DICOM SR file is generated and populated with this same clinical data. This SR file is written to the dataset directory. Please note that while these generated SR files can (and should) be made more structured, this is out of the scope of this functionality. Generated SR files have been tested and found to be compliant using DVTK.
- OnkoDICOM can now open SR files.
- If OnkoDICOM opens a clinical data SR file generated by OnkoDICOM, then the data within this SR file is used to populate the clinical data tab.

Please use this CSV and these datasets for testing:
- [CSV file](https://redmine.didymodesigns.com.au/attachments/737)
- [Dataset 1](https://redmine.didymodesigns.com.au/attachments/732)
- [Dataset 2](https://redmine.didymodesigns.com.au/attachments/739)
- [Dataset 3](https://redmine.didymodesigns.com.au/attachments/740)


JThomas2020 (changes requested):
Hey, mostly working, found one error with anonymisation:

`<class 'KeyError'> 'Comprehensive SR Storage' <traceback object at 0x7f11c054ad40>
  File "/home/josh/team_14/OnkoDICOM/src/Controller/ActionHandler.py", line 354, in anonymization_handler
    hashed_path = self.__main_page.call_class.run_anonymization(
  File "/home/josh/team_14/OnkoDICOM/src/Controller/MainPageController.py", line 397, in run_anonymization
    target_path = anonymize(path, dataset, filepaths, raw_dvh)
  File "/home/josh/team_14/OnkoDICOM/src/Model/Anon.py", line 811, in anonymize
    ds_pseudo_full_path = create_filename_from_dataset(
  File "/home/josh/team_14/OnkoDICOM/venv/lib/python3.8/site-packages/pymedphys/_dicom/anonymise/core.py", line 212, in create_filename_from_dataset
    mode_prefix = DICOM_SOP_CLASS_NAMES_MODE_PREFIXES[ds.SOPClassUID.name]
KeyError: 'Comprehensive SR Storage'`

To re-create:
1. Open Dataset 1 with the clinical data loaded
2. Click Anonymise

Also for the first time user window, I can't proceed if I only want to choose a csv file and no default directory, while it's probably not necessary as most people will specify a directory, would it be worthwhile to let a user save their csv data only?

I have a question with anonymization (this is probably out of scope so don't stress) but in Dataset 1, the clinical data contains gender, age and location info however those are blank in the DICOM tree, they're also blank in the row above the widget (i.e. Gender: one, Dob: ). Is anonymization meant to remove that information from the clinical data as well? And when anonymization occurs, will the SR file be stored? 

Otherwise, this is really great work! I've found no more issues and code looks neat with all relvant docstrings.

Tamookk:
Thank you for picking up the anonymisation problem, I'll get onto that right away. The first time window not letting you proceed is an oversight on my behalf too, I'll sort that out as well. 

> I have a question with anonymization (this is probably out of scope so don't stress) but in Dataset 1, the clinical data contains gender, age and location info however those are blank in the DICOM tree, they're also blank in the row above the widget (i.e. Gender: one, Dob: ). Is anonymization meant to remove that information from the clinical data as well?

I'm not actually sure. This never crossed my mind before. It would be difficult to do as the only constraint for the clinical data CSV is that the first column is the patient's ID. Without any strict standard or constraints it would be nearly impossible to determine which fields involve personal information without looking at the data and taking an educated guess (which in oncology I'd assume guessing is something you want to avoid), or prompting the user to ask them if any columns in the CSV contain identifying data. 

> And when anonymization occurs, will the SR file be stored?

This also never crossed my mind. Thinking about it, the SR files should be anonymised also during the anonymisation process as many attributes in the SR file are copied from other files in the dataset. However, doing this would be pointless if there was no way to remove identifying data from the clinical data stored in the SR. I'll ask around on what I should do concerning these two problems. Thank you for bringing them up, this is very important!

DuyTungHa:
The pipeline seems to run forever. (I have manually stopped it). 
Please fix this issue first then I will continue my review. 
I think it has something to do with the GUI testing.

Tamookk:
I had a look at the unit tests and noticed that they were stalling on the first time welcome window test due to the changes in the most recent commit. An error message dialogue is now appearing when it wasn't before, which the test was not changed to account for. I have updated the relevant unit test to hopefully fix this. Thank you for bringing this to my attention!

Tamookk:
I have been informed that tags in SR files should be anonymised as normal, but the anonymisation of the data within the file is up to the user. So, anonymisation has been fixed and now works with Comprehensive SR files.

JThomas2020 (commented):
When the anonymisation occurs, for some reason my Onko gets killed as the RAM approaches 8GB (the max ram for my VM). After killed, all files are anonymized however the DICOM-Sr file doesn't load. These are probably just be on my end though and I'll keep testing to try and pinpoint whats happening

Tamookk:
That's interesting. I just tested for myself now in both Windows and Ubuntu and RAM usage barely changed, if at all, when performing anonymisation. Please let me know if you find the problem!

JThomas2020:
Update: Testing on a fresh clone, anonymisation fails when opening a patent with data imported from clinicaldata.csv, not the SR file. SR file anonymisation works perfectly.

Tamookk:
Anonymisation works for me in this case but the generated SR file is not copied over to the anonymised directory, if that is what you mean? I'll fix that up.

JThomas2020:
Just to clarify, this is how I got the error

1. Load a patient who does not have an SR file but has data in the .csv, so clinical data is loaded from clinicaldata.csv
2. Immediately anonymise

Now when i reload the patient and choose to load the SR file as well, it all works for me. What i think is happening is while the information is there, because it's loaded via the csv, the sr file path is not stored in patient dict container (as it cannot have been selected before it exists). That's just my intuition

Tamookk:
Ah, I understand. I think what is happening is the patient dict container isn't updated with the SR when it is generated, so it was not being copied over to the anonymised dataset. I fixed that but encountered another problem - OnkoDICOM reads the series description for SR files to determine what data is stored in the SR (to see if it is clinical data and thus should be read and displayed), but this tag was being anonymised. So, when opening a dataset with an anonymised Clinical Data SR, OnkoDICOM didn't recognise it as such. Hopefully both of these problems have now been fixed!

JThomas2020 (approved):
Team 19, Great work guys!

src/Model/ROI.py
```diff
@@ -424,9 +424,16 @@ def get_pixluts(dict_ds):
...
+            if isinstance(ds, str):
+                if ds[0:3] != 'sr-':
+                    img_ds = dict_ds[ds]
+                    pixlut = calculate_matrix(img_ds)
+                    dict_pixluts[img_ds.SOPInstanceUID] = pixlut
+            else:
+                img_ds = dict_ds[ds]
+                pixlut = calculate_matrix(img_ds)
+                dict_pixluts[img_ds.SOPInstanceUID] = pixlut
+
```
DuyTungHa, line 436: I think you can merge the if statements. That will reduce the duplicated codes. (429-431 and 433-435).

DuyTungHa (commented):
I thought that this export clinical data function has been removed ?
image

src/Model/ROI.py
Tamookk: The code was written this way under the assumption that if the `if` statements are merged, e.g. `if (isinstance(ds, str)) and (ds[0:3] == 'sr-')`, and ds wasn't a string, then there would be a type error with the second operand. I did this as I wasn't sure if Python checked both operands, or just the first. However upon getting this comment I had a look at the Python documentation, which states:

> The expression `x and y` first evaluates _x_; if _x_ is false, its value is returned; otherwise, _y_ is evaluated and the resulting value is returned.

https://docs.python.org/3/reference/expressions.html?highlight=boolean#boolean-operations

So I think the if statements can safely be merged. Thank you for bringing this up! Thanks for bringing up the "Export Clinical Data" option in the toolbar too. Just a simple oversight that will also be fixed.

minhhuyxx:
Hey, I notice that several type-1 tags are empty. Have you discussed this with Dr Swerdloff?

Tamookk:
All tags that are empty in the code are ones that are required but can be left empty if unknown. As stated in the description for the PR, these generated SR files can (and should) be made more structured (which includes filling out these tags, structuring the data better than a single text dump, etc.), but all of this is out of the scope of this functionality. It's expected that future teams/development could build upon this work to achieve this, as the aim of this functionality was to get the bare minimum required to generate a compliant file (compliance has been tested with DVTK and the files pass) and fill it with clinical data.

SR file generation was done in consultation with the resident DICOM expert who himself stated that doing this properly is very complicated and would take quite a while to achieve. It was recommended that it should be improved, however due to the limited amount of time available for development on this functionality it was agreed that it is fine as-is. I hope this answers your question!

(and I hope I haven't misunderstood anything during my SR generation discussions and done something majorly wrong here!)

DuyTungHa (approved):
Team23

PR #168, opened 2021-09-21
Hierarchy of objects

minhhuyxx:
This PR changes the original DICOM structure tree into a hierarchy of objects which is required to distinguish RT files if there are more than one RT file of each type existing in the same directory. Please refers to the original [comment](https://github.com/didymo/OnkoDICOM/pull/126#issuecomment-885979975) about the hierarchy of objects if you need more info about it.

- New UI for open patient window's DICOM tree: Image series are always displayed upon searching completed. Users can open/close a node's child branches by clicking on the node. 
- New file selecting logic: RT files can only open with an image set if the RT files reference the image set or reference another RT file that references the image set. SeriesInstanceUID, SOPInstanceUID, FrameOfReferenceUID, and ReferencedFrameOfReferenceUID are used to check the referencing between files. All other logics are kept the same (only one file of each type can be selected, only one patient can be selected, etc.)
- Popup window to select a RTSS from a list of existing RTSSs attached to the selected image set: When users only open an image set, then proceed to draw new ROIs or convert isodoses to ROIs, the program will pop up a window to prompt for an exact RTSS that the users want the new ROIs to be saved to from a list of existing RTSSs.


src/View/util/SelectRTSSPopUp.py
```diff
@@ -0,0 +1,122 @@
...
+        # Add all components to a vertical layout
+        self.layout = QVBoxLayout()
+        self.layout.addWidget(self.explanation_text)
+        self.layout.addWidget(self.scroll_area)
+        self.layout.addWidget(self.button_area)
+        self.setLayout(self.layout)
+
+    def on_checkbox_clicked(self, checkbox):
+        """
+        function to control selected rtss
```
Tamookk, line 103: Missing `:param checkbox` here.

Tamookk (commented):
Everything works as described, no issues encountered with new functionality. Fantastic work! Just that one small thing to fix up and I'll approve.

src/View/util/SelectRTSSPopUp.py
minhhuyxx: Thanks, @Tamookk. Comment has been added to clarify that function.

Tamookk (approved):
Great work! Team 14

JThomas2020 (approved):
Team 19, great job, everything works as described

PR #169, opened 2021-09-21
Add SUV2ROI Functionality

Tamookk:
This PR adds SUV to ROI functionality to OnkoDICOM. An overview of features added:

- When a PET dataset has been opened, a new button appears in the bottom right of the DICOM View that allows the user to initiate the SUV2ROI process.
- When clicking this button, PET pixel values (stored in units of Bq/mL) are converted into standard uptake values (SUVs). These SUVs are then used to generate ROIs for each SUV level, from 1 up to the maximum SUV value.
- Patient weight is required for the SUV2ROI process, so if a dataset does not include the patient's weight, the user will be prompted to enter it in kilograms. If they do not enter a number, they will continue to be prompted until they do, or close the prompt. If no patient weight is entered, SUV2ROI does not take place.

Please note that SUV to ROI functionality only works on PET files that are both attenuation and decay corrected - the process will stop and not proceed otherwise (however from the PET datasets I have seen, every one of them was attenuation and decay corrected).

Please use the following datasets for testing: [dataset 1](https://redmine.didymodesigns.com.au/attachments/731), [dataset 2](https://redmine.didymodesigns.com.au/attachments/733), [dataset 3](https://redmine.didymodesigns.com.au/attachments/713). None of these datasets include patient weight, so when prompted for patient weight enter a number in the range of 70-80.


JThomas2020 (commented):
Great work guys, one question, if a patient already has a weight inputted, is it possible to retrieve the stored patent weight when the patient weight dialog opens again, might be good if someone forgets the weight they inputted?

Tamookk:
That's a good idea, however I'm not sure how often the scenario would arise where SUV2ROI would have to be performed on the same dataset twice. Unless the user deletes the RT Struct or removes all SUV ROIs by accident, for example.

JThomas2020:
Thats fair, if thats the case then would it be worthwhile to hide the button if a weight is already stored?

Tamookk:
The weight is currently not stored after it's entered. There's nowhere to store it (except perhaps in the PET DICOM files), however I don't want to add anything to these in case the weight is inaccurate. While it is a bit inconvenient, I think that leaving the button and asking for a weight every time (if it does not already exist in the PET files) is the safest route.

JThomas2020 (approved):
Team 19, Great work guys, thanks for being quick to respond to my questions

DuyTungHa:
I have a few comments:
1. If you open dataset2, series [WB_NAC] (255 images), and use the function with the input of 80kg, in the terminal you will see "Boundaries could not be calculated" and the following window will pop-up. Is this situation expected ? 
image
2. I think that you should use onkoDICOM's stylesheet for your window/dialog. Currently, the buttons are using default styling of PySide6.

Tamookk:
Interesting regarding the pop-up. That isn't expected, I'm not sure why that's triggering. I'll see if I can find the cause. As for the styling, I'll fix that up.

DuyTungHa (commented):
I have some other comments:
1. Should we have a stricter input validation? I mean I can enter a negative integer value and the process still runs. 
2. If the input value is invalid, I think it is better to disable the ok button and show an indicator underneath the input bar. (This one is just a recommendation. Up to you whether you want to do it).

Tamookk:
Oops, I forgot about negative numbers... That's fixed. I'll leave the feedback for invalid inputs for now. What is there works and it's easy enough to change in the future if needed.

DuyTungHa (approved):
Team23

JThomas2020 (approved):
Team 19, just re-confirming approval. Good job guys!

PR #170, opened 2021-09-22
Remove file selecting restriction

minhhuyxx:
This PR partially removes some file selecting restrictions. Now, users can open RT files as long as they are all attached to the same image set.

This PR also fixes the SelectRTSSPopup window focus issue. After running ISO2ROI, the popup window used to lose focus and hide behind the main window.


Tamookk (approved):
Works as described, good work.
Team 14

JThomas2020 (approved):
Team 19, great work guys!

PR #171, opened 2021-09-26
UX Improvements

Tamookk:
This PR makes multiple user experience improvements. These improvements are:

- A reason for failure is now given if the SUV2ROI process fails.
- The footer on the main page has been updated to read "@OnkoDICOM2021".
- The clinical data CSV page in the add-on options menu has had its text updated.
- "RT DOSE" in the calculate DVH prompt that appears when OnkoDICOM loads a patient has been changed to "RTDOSE".

### DIFF
--- a/src/View/mainpage/MainPage.py
+++ b/src/View/mainpage/MainPage.py
@@ -220,7 +220,7 @@ class UIMainWindow:
         layout_footer = QtWidgets.QHBoxLayout(self.footer)
         layout_footer.setContentsMargins(0, 0, 0, 0)
 
-        label_footer = QtWidgets.QLabel("@OnkoDICOM 2019-20")
+        label_footer = QtWidgets.QLabel("@OnkoDICOM2021")
         label_footer.setAlignment(QtCore.Qt.AlignRight | QtCore.Qt.AlignRight)
 
         layout_footer.addWidget(label_footer)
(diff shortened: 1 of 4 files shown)


JThomas2020:
Looks good guys, just wondering, is it possible to have multiple reasons for failure? If so then you could store an an array

Tamookk:
Due to the way SUV2ROI is done, as soon as one failure condition is satisfied the process will return. The first failure that causes this is the only one recorded for this reason. There's no need to continue to step 2 if step 1 fails, in other words, so we don't know if step 2 will fail if step 1 did!

JThomas2020 (approved):
Team 19, Good Job!

src/View/mainpage/MainPage.py
```diff
@@ -220,7 +220,7 @@ def create_footer(self):
         layout_footer = QtWidgets.QHBoxLayout(self.footer)
         layout_footer.setContentsMargins(0, 0, 0, 0)
 
-        label_footer = QtWidgets.QLabel("@OnkoDICOM 2019-20")
+        label_footer = QtWidgets.QLabel("@OnkoDICOM2021")
```
tdlinh2712, line 223: shouldn't it be "@OnkoDICOM 2021"? I'm talking about the space between the word and number, that's it

tdlinh2712 (approved):
just have that one question about the space, the rest are all good to me! team 23

src/View/mainpage/MainPage.py
Tamookk: As it is in the PR, with no space, is exactly how the product owner requested it in Slack. This was also shown off during our recent meeting and no complaints were raised!

tdlinh2712: all good then!

PR #172, opened 2021-09-26
Bug: Normalising Image data for correct viewing brightness

JThomas2020:
This PR fixes bugs with images being too dark or having incorrectly scaled units by using the rescale data stored in DICOM files. Specifically, this PR:
- applies a linear transform to the incoming DICOM data to get it in a standardised viewing scale
- transforms CT data from CT# to Hounsfield, and translating so all units are positive
- corrects the use of the window center and window width data, to load an initial window/level that is usable
- translates fusion output so all values are positive

When testing, please test with different images, things to look out for while testing:
- Transect data is within the appropriate ranges, no data should be less than 0
- For CT: windowing shows the appropriate regions
- Fused images are as bright as the normal view


sjswerdloff:
The scaling you are applying to CT data is *to* Hounsfield units, which are the normalisation of attenuation coefficients relative to that of water for a given set of scan parameters, xray subsystem and detection subsystem.

didymo:
> The scaling you are applying to CT data is _to_ Hounsfield units, which are the normalisation of attenuation coefficients relative to that of water for a given set of scan parameters, xray subsystem and detection subsystem.

Should different units be used for different scan types? What would you do in this case?

JThomas2020:
> The scaling you are applying to CT data is _to_ Hounsfield units, which are the normalisation of attenuation coefficients relative to that of water for a given set of scan parameters, xray subsystem and detection subsystem.

You’re right and I should clarify, the CT data is now scaled to Hounsfield units, however Andrew wants all units to be greater than 0 as images already in Hounsfield are too dark.

src/Model/MovingModel.py
```diff
@@ -76,7 +80,16 @@ def create_moving_model():
 
     moving_dict_container.set("dict_windowing_moving", dict_windowing)
 
-    pixel_values = convert_raw_data(dataset)
+    slope = 1
+    intercept = 0
+
+    if not moving_dict_container.has_attribute("scaled"):
```
Tamookk, line 86: Don't quite understand this block of code here. The if/else does nothing as `pixel_values` is set again right under this block, meaning anything in the if/else is discarded and unused.

Also, `is_pt` is the rescale intercept, right? Why hard code this value instead of grabbing it from the DICOM file? And then if this is true, why is the rescale intercept stored in both `is_pt` and `intercept` but with potentially different values?

Also a little confused about the rescale slope, which is specified as the integer 1 but passed into `convert_raw_data` as a boolean.

The name of `is_pt` is a bit misleading too, as it implies this would be a boolean when it is in fact an integer.

The code all works fine, I just think my confusion comes down to the variable names.

Tamookk (commented):
Code works as described in the PR description. All PEP8 compliant. Just two questions.

When performing a transect, some data (eg black area surrounding the patient) appears as less than 0. Is this expected despite the PR description stating "[test that] Transect data is within the appropriate ranges, no data should be less than 0"?

sjswerdloff:
> > The scaling you are applying to CT data is _to_ Hounsfield units, which are the normalisation of attenuation coefficients relative to that of water for a given set of scan parameters, xray subsystem and detection subsystem.
> 
> Should different units be used for different scan types? What would you do in this case?

Hounsfield Units are the appropriate scale to use for all CT.
They pin the attenuation of air to -1000 and water to 0.
Different energy CT will result in somewhat different values for materials with different average atomic number.
So you can't directly compare two CT that have different kVp (kilo-Voltate Peak) and different filtration (typically a thin metal layer to avoid having a lot of low energy xrays that get fully absorbed in the body and don't provide measurable signal).

The presentation layer can use whatever scaling you like (e.g. -1000 to +3000) to address the fact that some of the body (lung in. particular) is less xray dense/opaque (attenuation is the proper physical parameter) than water.

Spectral or Dual Energy CT is the modern approach to helping distinguish the atomic components and have a more quantitative understanding of the underlying tissue,  but for ease of use for existing software an algorithms, the additional information is reduced to create a pseudo MonoEnergetic CT (where typical CT is polyenergetic but convolved in measurement so the information is lost).

sjswerdloff:
> > The scaling you are applying to CT data is _to_ Hounsfield units, which are the normalisation of attenuation coefficients relative to that of water for a given set of scan parameters, xray subsystem and detection subsystem.
> 
> You’re right and I should clarify, the CT data is now scaled to Hounsfield units, however Andrew wants all units to be greater than 0 as images already in Hounsfield are too dark.

You need to distinguish between presentation being scaled (window level) and underlying data being scaled.
If you want to have your underlying data to be scaled to Hounsfield + 1000, that's OK, but it needs to be clear to the user (at least in documentation if not on-screen) what the underlying data has been scaled to.  Just be very careful if you write out CT data to make sure that the appropriate scaling values *for the pixel data* are encoded.

src/Model/MovingModel.py
sjswerdloff: I have the same confusion with is_pt

I think it's really important to distinguish between scaling of the underlying data and rescaling for presentation.

JThomas2020: > Also, is_pt is the rescale intercept, right? Why hard code this value instead of grabbing it from the DICOM file? And then if this is true, why is the rescale intercept stored in both is_pt and intercept but with potentially different values?

This is a bit confusing so I will also clarify and use examples: 
- in [DICOM-RT-02](https://onkodicom.com.au/sites/default/files/2020-01/DICOM-RT-02.zip) the scans have a rescale slope of 1 and rescale intercept of -1024, however the image itself is already bright enough for Andrew, so when applying the rescale, I add 1024 so 1024-1024=0 i.e. the intercept does not change. 
- in [LUNG1-010](https://redmine.didymodesigns.com.au/attachments/762) we have a rescale slope of 1 and an intercept of 0, meaning this data is already in Hounsfield, however [Andrew's comment](https://onkoworkspace.slack.com/archives/CGRSJ6H16/p1631838677111500) means this is too dark. Testing i found increasing the intercept by 1024 to be the best way to do it. Another way is that DICOM files store the maximum bit depth, which does work, however
- in [test image fusion dataset](https://redmine.didymodesigns.com.au/attachments/744), the slope is 1, and the rescale is 0, but almost all the data is greater than -1024, except for the endges which have a value of -2048, the maximum bit depth here because of this is -2048 and using this to adjust the intercept will make every other value brighter than recorded. This does mean there will be values less than 0 but that is due to the data coming in

So the reason there is an is_pt is because there is no good way in DICOM to rescale out of Hounsfield, especially if the image is already in hounsfield.

I also just realised this is meant to be for CT, not PT, as PT does not need this intercept change, so I'll fix this

JThomas2020: All this data is accessablie in thr Dicom Tree tab

JThomas2020:
@sjswerdloff Thanks for all the information. All the scaling values are not changed in the DICOM files, so if a user was to open up the DICOM Tree tab, they would see the values that have been used for scaling in the tree. For rescaling the PT, do you think its appropriate to separate the + 1000 out of the dataset rescale and apply it to generating the images? That way data from performing the transect would be in Hounsfield however the files would work with all the current existing windowing options.

JThomas2020:
> When performing a transect, some data (eg black area surrounding the patient) appears as less than 0. Is this expected despite the PR description stating "[test that] Transect data is within the appropriate ranges, no data should be less than 0"?

This is expected, and I can't do much about that as I'd rather not apply any non-linear, non-reversible transformations to the data. As long as everything in the circle around the patient and all the windowing works as intended, its fine.

src/Model/MovingModel.py
JThomas2020: > Don't quite understand this block of code here. The if/else does nothing as pixel_values is set again right under this block, meaning anything in the if/else is discarded and unused.

This was a mistake and has been fixed, thanks for the pickup.

Tamookk: I think I understand now. Thank you for the explanation, it was very helpful!

Tamookk (approved):
Thanks for fixing that up. All works well, good work! Team 14

src/Model/ImageFusion.py
```diff
@@ -347,7 +347,7 @@ def get_fused_window(level, window):
     color_sagittal = {}
     color_coronal = {}
 
-    windowing = (int(level), int(window))
+    windowing = (int(level-1024), int(window))
```
tdlinh2712, line 350: Please chang

tdlinh2712, line 350: Please change the "magic number" 1024 here to the constant CT_RESCALE_INTERCEPT

JThomas2020: Missed this, thanks and done!

tdlinh2712 (approved):
Works well on my end! Team 23.

PR #173, opened 2021-09-26
Fix: Image Fusion Window

peteqian:
This PR resolves bug [#1922](https://redmine.didymodesigns.com.au/issues/1922) and [#1935](https://redmine.didymodesigns.com.au/issues/1935)

- Fix the user being able to fuse with the same image/series. 
- Fix when attempting image fusion with a new patient after already performing fusion with the previous patient, the window should now be able to detect the new patient as the current patient while selecting images and not the previous patient.


src/View/ImageFusion/ImageFusionWindow.py
```diff
@@ -284,19 +278,22 @@ def retranslate_ui(self, open_image_fusion_select_instance):
         self.open_patient_window_confirm_button.setText(_translate(
             "OpenPatientWindowInstance", "Confirm"))
 
-    def update_new_patient(self):
+    def update_patient(self):
+        # self.clear_checked_leaves()
```
Tamookk, line 282: This line should be removed if it is unused.

src/View/ImageFusion/ImageFusionWindow.py
```diff
@@ -284,19 +278,22 @@ def retranslate_ui(self, open_image_fusion_select_instance):
...
         self.patient_dict_container = PatientDictContainer()
         self.patient = self.patient_dict_container.get("basic_info")
         self.patient_id = self.patient['id']
-        self.clear_checked_leaves()
+
+        dataset = self.patient_dict_container.dataset[0]
+        self.patient_current_image_series_uid = dataset.get("SeriesInstanceUID")
```
Tamookk, line 288: Slightly too long for PEP8 compliance.

src/View/ImageFusion/ImageFusionWindow.py
```diff
@@ -350,105 +347,173 @@ def search_progress(self, progress_update):
...
-        if dicom_structure is None:  # dicom_structure will be None if
-            # function was interrupted.
+        # dicom_structure will be None if function was interrupted.
+        if dicom_structure is None:
             return
 
         for patient_item in dicom_structure.get_tree_items_list():
-            self.open_patient_window_patients_tree.addTopLevelItem(
-                patient_item)
+            self.open_patient_window_patients_tree.addTopLevelItem(patient_item)
```
Tamookk, line 368: Slightly too long for PEP8 compliance.

src/View/ImageFusion/ImageFusionWindow.py
```diff
@@ -350,105 +347,173 @@ def search_progress(self, progress_update):
...
-
-        # Check the existence of IMAGE, RTSTRUCT and RTDOSE files
-        if len(list({'CT', 'MR', 'PT'} & selected_series_types)) == 0:
-            header = "Cannot proceed without an image file."
-            self.open_patient_window_confirm_button.setDisabled(True)
+        # Check selected items and display warning messages
+        self.check_selected_items(selected_patient)
+
+    def display_a_tree_branch(self, node, is_expanded):
+        """ Expand/Close a tree branch """
```
Tamookk, line 425: Docstring is incomplete, parameter definitions are needed.

src/View/OpenPatientWindow.py
```diff
@@ -146,7 +146,7 @@ def setup_ui(self, open_patient_window_instance):
             self.tree_item_clicked)
         self.open_patient_window_instance_vertical_box.addWidget(
             self.open_patient_window_patients_tree)
-        self.last_patient = None
+
```
Tamookk, line 150: Too many blank lines

Tamookk (commented):
All works as described. Just a few PEP8 compliance issues to correct and I will approve.

src/View/ImageFusion/ImageFusionWindow.py
peteqian: Could Team 23 advise what is 'is_expanded' meant to be?

Tamookk: I can't say for sure but it looks like this variable is used to set whether a node on the tree (when selecting a patient/image set for image fusion) is expanded on not.

peteqian: I didn't write this code. I can't comment - I also don't want to assume. Can team 23 please fix their docstring as well as their own function within the next PR, thanks.

Tamookk (approved):
Great work, all works as described and is PEP8 compliant (assuming the docstring will be fixed in the future!). Team 14

src/View/ImageFusion/ImageFusionWindow.py
tdlinh2712: @peteqian i think this is from last year's team... Regardless, it is a boolean parameter to mark whether a node 's information is expanded. If you can update the docstring here for us would be great. Thank you.

peteqian: This isn't last year's code. This code was developed by your team in this [PR](https://github.com/didymo/OnkoDICOM/pull/168/files). I will make the fix regarding my file, however, upon investigating this flag and how it sets - shorthand notation makes it difficult to read if the flag is constantly set to False or True due to the PEP-8 Compliance.

`
is_expanded = False \
            if (item.isExpanded() is True and
                item.checkState(0) == Qt.CheckState.Unchecked) else True
        self.display_a_tree_branch(item, is_expanded)
`

Also, a suggestion might be to is to remove the "is True" section of the if condition, as this makes the condition require an additional useless parameter for evaluating Truths.

tdlinh2712: Yep, thanks for pointing that out, and sorry I thought it was last year's code 😂 we will fix the condition in our next PR so that it's more readable.

tdlinh2712: @peteqian that being said, the code in here is copied from PR #168. It's a bit concerning since it shows that from now on, if we make changes to Open Patient Window or how the directory tree is structured, we need to do the same thing for the one In Image Fusion as well. It makes it more difficult to maintain the code. I'm not saying that we need to fix the repetition now, but our team will document it for next year's aims planning so that they can work on integrating these two windows together.

peteqian: But does thou approve such a request?

tdlinh2712: @peteqian i'm still testing the program for functionalities, so if it all works well i'll approve this PR for sure

peteqian: > @peteqian that being said, the code in here is copied from PR #168, is that correct? It's a bit concerning since it shows that from now on, if we make changes to Open Patient Window or how the directory tree is structured, we need to do the same thing for the one In Image Fusion as well. It makes it more difficult to maintain the code. I'm not saying that we need to fix the repetition now given the time constraint, but our team will document it for next year's aims planning so that they can work on integrating these two windows together.

I do agree with this. However, previously when I was trying to inherit this class - I was running into issues where the subclass (UIImageFusionWindow) was not instantiating the superclass (UIOpenPatientWindow) properly - this was most likely due to no __init__() present in the UIOpenPatientWindow class (although just speculating at this point). If you were to make the class inheritable - then refactoring would be very easy.

tdlinh2712 (approved):
works well! team 23

PR #178, opened 2021-10-01
Add Batch SUV2ROI

Tamookk:
This PR adds batch processing SUV2ROI:

- It adds another tab to the batch processing window and allows the user to perform batch processing SUV2ROI.
- Inside this tab the user is presented with a message on PET files that are compatible with SUV2ROI.
- The user is also shown a table of patient IDs and entry fields. This allows the user to enter in patient weights if they do not exist in the datasets. These fields are checked to ensure that they are positive numbers (please note that messages on incorrect entries to these fields are not presented to the user until the summary window).
- The batch processing summary screen now includes a summary on SUV2ROI functionality.


minhhuyxx:
Team 23, great work guys! Everything works as described. I just have one suggestion. You might consider setting a validator for the fields using a regular expression. This will ensure users can only put down correct values. Please refers to DrawROI's input field for alpha value for an example.

Tamookk:
A check already exists to ensure that numbers entered into fields are non-negative numbers, however this is not enforced and the user is only informed in the summary window. This check does not enforce that the fields are filled however, as the user may not have the patient weight available. I have made a commit that now shows a warning dialog box if entered weights are non-compliant, which is similar behaviour to the "Enter Patient Weight" dialog box for individual patient curation. Thank you for the suggestion.

src/Controller/BatchProcessingController.py
```diff
@@ -191,6 +201,61 @@ def perform_processes(self, interrupt_flag, progress_callback):
                 self.batch_summary[patient]["iso2roi"] = reason
                 progress_callback.emit(("Completed ISO2ROI", 100))
 
+            # Perform SUV2ROI on patient
+            if 'suv2roi' in self.processes:
```
tdlinh2712, line 205: I don't think using if... else here to decide the operations to do is maintainable code. Since there will be more values in `self.processes`, the if/else thing makes this `perform_processes` function really long. Can you put the operations for each case in separate functions then call it using a dictionary/class of functions? (There are many suggestions on how to do it here: https://stackoverflow.com/questions/9168340/using-a-dictionary-to-select-function-to-execute)

Tamookk: Thank you for this suggestion! It this is okay, it will be left alone for this PR but will be changed in the next one.

tdlinh2712: Sure, as long as they're fixed!

Tamookk: Of course, I'm working on it asap in preparation for the next PR.

JThomas2020 (approved):
Team 19, Great Work! Keen for the next one

minhhuyxx (approved):
Team 23. Sorry for the delay.

PR #180, opened 2021-10-02
Add Batch PyRad2CSV

Tamookk:
This PR adds batch processing of exporting Pyradiomics data to a CSV file:

- This PR adds another tab to the batch processing window which allows the user to select the output CSV file path for PyRad2CSV.
- The functionality performs Pyrad2CSV on all patients in the selected folder, and saved the radiomics data into a single CSV file.
- The batch processing summary screen now includes a summary on PyRad2CSV functionality.
- The Github test has been updated to install plastimatch so that Pyrad2CSV can be properly tested.


sjswerdloff:
Is this only intended for *nix environment s?
The commands invoked are based on strings that appear to be  *nix (e.g 'rm')

It's OK if this is expected to be os constrained.
Otherwise the python platform agnostic commands (potentially imported from a package) should be used.

Tamookk:
This is not intended to be OS constrained, and it looks as though if you're not running a *nix OS then the delete command (the only one that would differ between platforms) is replaced with a Windows-specific one. I've been testing this code on Windows and everything seems to work okay, with Pyradiomics running properly and temporary files and folders being removed.

Using Python commands (e.g. `os.remove`) is definitely better for long-term maintainability, but if it ain't broke... 

The code that runs these commands in this PR was copied from the existing Pyradiomics code, also. We didn't want to change too much here in case something broke, somewhere.

src/Controller/BatchProcessingController.py
```diff
@@ -320,6 +325,38 @@ def batch_dvh2csv_handler(self, interrupt_flag,
...
+        Handles creating, starting, and processing the results of batch
+        Pyrad2CSV.
+        :param interrupt_flag: A threading.Event() object that tells the
+                               function to stop loading.
+        :param progress_callback: A signal that receives the current
+                                  progress of the loading.
+        :param patient: The patient to perform this process on.
+        """
+        # Get current files
+        cur_patient_files = self.get_patient_files(patient)
```
eddieGong2604, line 340: Same problem as previous PR here. Static method should not be called by an instance

src/Model/batchprocessing/BatchProcessPyRad2CSV.py
```diff
@@ -0,0 +1,287 @@
...
+            self.summary = "INTERRUPT"
+            return False
+
+        if not self.ready:
+            self.summary = "SKIP"
+            return False
+
+        rtss_path = self.patient_dict_container.filepaths.get('rtss')
+        patient_id = self.patient_dict_container.dataset.get('rtss').PatientID
+        patient_id = self.clean_patient_id(patient_id)
```
eddieGong2604, line 67: clean_patient_id is static method, should not be called by an instance

eddieGong2604 (commented):
Please fix the problem of static method getting called by instance

src/Controller/BatchProcessingController.py
Tamookk: These have been fixed. Thank you for picking these up! Must have missed them before I created the PR.

src/Controller/BatchProcessingController.py
```diff
@@ -320,6 +325,39 @@ def batch_dvh2csv_handler(self, interrupt_flag,
...
+        process = BatchProcessPyRad2CSV(progress_callback,
+                                        interrupt_flag,
+                                        cur_patient_files,
+                                        self.pyrad_output_path)
+        process.set_filename('PyRadiomics_' + self.timestamp + '.csv')
+        success = process.start()
+
+        # Set summary message
+        if success:
+            reason = "SUCCESS"
```
eddieGong2604, line 351: If you can make summary messages enum or constants, and import them when you want to use that would be cleaner.

src/Model/batchprocessing/BatchProcessPyRad2CSV.py
```diff
@@ -0,0 +1,290 @@
...
+    def start(self):
+        """
+        Goes through the steps of the PyRadCSV conversion.
+        """
+        # Stop loading
+        if self.interrupt_flag.is_set():
+            self.patient_dict_container.clear()
+            self.summary = "INTERRUPT"
+            return False
+
```
eddieGong2604, line 60: If you can make summary messages enum or constants, and import them when you want to use that would be cleaner.

JThomas2020 (approved):
Team 19, sorry for the delay, great job guys! everything works and code is high quality

src/Model/batchprocessing/BatchProcessPyRad2CSV.py
```diff
@@ -0,0 +1,290 @@
...
+            patient_path, patient_id, patient_nrrd_file_path, mask_folder_path)
+
+        # Stop loading
+        if self.interrupt_flag.is_set():
+            self.patient_dict_container.clear()
+            self.summary = "INTERRUPT"
+            return False
+
+        if radiomics_df is None:
+            self.summary = "PYRAD_NO_DF"
```
eddieGong2604, line 128: If you can make summary messages enum or constants, and import them when you want to use that would be cleaner.

eddieGong2604 (commented):
Team 23: Please consider making possible summary values constants instead of hardcode.
I'm still waiting for the pyrad2CSV to finish on my machine. How long does it normally take?

Tamookk:
I'm not sure how well making summary values constants would work for some later PRs we are making, as some specific information is directly contained in these summary strings (such as numbers that need to be extracted). This information can't be passed over if the summary values are made constant. I will definitely look into this however, but not for this PR.

The amount of time Pyrad2CSV takes depends on how many datasets you perform it on, and how many ROIs are in these datasets. The more datasets and ROIs, the longer it takes. That is not due to how we coded this PR as the actual radiomics calculations are done by external program. My suggestion for testing would be to perform it on one dataset with a small amount of ROIs. The longest I have seen it take personally is about half an hour for 5 datasets with many large ROIs in them.

eddieGong2604 (approved):
Team 23 approves this PR

PR #184, opened 2021-10-03
Add Batch ROI Name Cleaning and ROI Name to FMA ID

Tamookk:
This PR adds both Batch ROI Name Cleaning and ROI Name to FMA ID conversion. Two tabs are added, one for each process. For batch ROI Name Cleaning:

- The tab for Batch ROI Name Cleaning includes a table which lists non-standard ROI names, options to either ignore, modify, or delete the ROI, the new ROI name, and a drop-down list displaying the datasets that this ROI name was found in.
- Non-standard names are ROI names that are not found in `/data/csv/organName.csv`, or do not have prefixes found in `/data/csv/volumeName.csv`.
- If an ROI name is in the standard organ name file, but in a different case, then the default option is set to "Modify", and the default new name is set to the equivalent standard name.
- If an ROI name has the correct prefix but with the wrong case, then a suggested new name is generated instead of a drop-down list of organ names.
- A summary of ROIs, what datasets they belong to, and what was done to them has been added to the batch summary window for Batch ROI Name Cleaning.

For batch ROI Name to FMA ID:

- The tab for Batch ROI Name to FMA ID includes a short amount of text explaining the process.
- ROIs are automatically renamed to the equivalent FMA ID if the ROI name is found in `/data/csv/organName.csv`.
- The amount of ROIs renamed into their FMA IDs for each patient is added to the batch summary window.


JThomas2020 (approved):
Team 19, Great work overall guys!

eddieGong2604 (approved):
Team 23 approved this PR. Great work.

PR #185, opened 2021-10-03
PET-CT Fusion Tab

JThomas2020:
This PR adds a PET-CT viewer to OnkoDICOM. This allows a PET and CT Image to be loaded and overlayed for comparison.

Functionality added:
- Viewer which displays both PET and CT
- Alpha slider to determine the mix of PET and CT
- Windoing to both PET and CT
- conversion of greyscale image sets to colored image sets (In this PR, PET is displayed as a heat map)

How to use:
1. Install open cv: `pip install opencv-python`
2. Make sure you have a PET/CT image set, if you dont, one can be found [here](https://redmine.didymodesigns.com.au/attachments/716)
3. Open any image set (it doesn't need to be PET or CT)
4. Navigate to the PET CT tab (on the right of dicom view)
5. Press the button and select both the PET and CT image


Tamookk:
The viewer works well for all 3 views, and the alpha slider works just as I would expect. Windowing also works perfectly, changing in both the DICOM View and PET/CT View. Great work there!

However I am a little confused regarding the colouring/heat map. For PET images, whiter areas (such as the brain, bladder, and kidneys) are areas where the most activity was detected during a PET scan, and so are areas I would expect be coloured in. However in this PR it appears as though the opposite is true, and these areas aren't coloured in as much (I would expect these areas to be "hot", coloured red). Is this intended/what the product owner requested? (Bear in mind I'm not a doctor so my expectations could be wildly wrong!)

Also, am I supposed to be able to perform PET/CT fusion on Patient B while I have Patient A opened? Because I currently can do this and it doesn't make much sense to me that I can do PET/CT fusion on a patient that is different to what I already have opened (but again, not a doctor, so feel free to tell me if/why I am wrong).

I will review the code in-depth tomorrow, but what I have seen so far is of high quality. This functionality is a great achievement!

JThomas2020:
> However I am a little confused regarding the colouring/heat map. For PET images, whiter areas (such as the brain, bladder, and kidneys) are areas where the most activity was detected during a PET scan, and so are areas I would expect be coloured in. However in this PR it appears as though the opposite is true, and these areas aren't coloured in as much (I would expect these areas to be "hot", coloured red). Is this intended/what the product owner requested? (Bear in mind I'm not a doctor so my expectations could be wildly wrong!)

Yes this is intended. At Andrew's request, the heat transformation goes Black --> Red --> Yellow --> White. The heat map converts the greyscale input of the PET scan into this color scheme. We're using open-cv's mapping function to do this and the map itself can be easily changed. Thanks for checking and I should've clarified this in the PR

> Also, am I supposed to be able to perform PET/CT fusion on Patient B while I have Patient A opened? Because I currently can do this and it doesn't make much sense to me that I can do PET/CT fusion on a patient that is different to what I already have opened (but again, not a doctor, so feel free to tell me if/why I am wrong).

Yes, the PET/CT viewer can open different patients. This is because PET/CT has its own dict_container that identifies and separates the CT and PET images and is meant to be separate from patient_dict_container. Originally I wrote this so it would operate similarly to image fusion (working off the current patient), however Andrew specified that both a PET and a CT should be selected at the same time. I'll check with Andrew to see if he wants the patient to be the same, but im certain the image sets selected can be different.

peteqian:
> > However I am a little confused regarding the colouring/heat map. For PET images, whiter areas (such as the brain, bladder, and kidneys) are areas where the most activity was detected during a PET scan, and so are areas I would expect be coloured in. However in this PR it appears as though the opposite is true, and these areas aren't coloured in as much (I would expect these areas to be "hot", coloured red). Is this intended/what the product owner requested? (Bear in mind I'm not a doctor so my expectations could be wildly wrong!)
> 
> Yes this is intended. At Andrew's request, the heat transformation goes Black --> Red --> Yellow --> White. The heat map converts the greyscale input of the PET scan into this color scheme. We're using open-cv's mapping function to do this and the map itself can be easily changed. Thanks for checking and I should've clarified this in the PR
> 
> > Also, am I supposed to be able to perform PET/CT fusion on Patient B while I have Patient A opened? Because I currently can do this and it doesn't make much sense to me that I can do PET/CT fusion on a patient that is different to what I already have opened (but again, not a doctor, so feel free to tell me if/why I am wrong).
> 
> Yes, the PET/CT viewer can open different patients. This is because PET/CT has its own dict_container that identifies and separates the CT and PET images and is meant to be separate from patient_dict_container. Originally I wrote this so it would operate similarly to image fusion (working off the current patient), however Andrew specified that both a PET and a CT should be selected at the same time. I'll check with Andrew to see if he wants the patient to be the same, but im certain the image sets selected can be different.

It would be worth noting that color maps such as magma or inferno contrast from black to white via purple to yellow, purple being what represents "cooler" or lower intensity pixels and as intensities become more "hot", it transitions to white via yellow.

tdlinh2712 (commented):
Functionality-wise, it works well, congrats!
Got one question regarding the architecture: Why do we need another class of PETCTDictContainer? From what I see when testing, the PET/CT viewer tab doesn't use anything in the PatientDictContainer, is that correct? If that's the case, why does the user need to open a non-related image before being able to open a PET/CT image? Why cant they open the PET/CT in the first open patient window? Not to mention the code repetition of having a third DictContainer class.

src/Controller/GUIController.py
```diff
@@ -105,7 +108,23 @@ def update_ui(self):
...
-        
+
+    def open_patient(self, progress_window):
+        self.go_next_window.emit(progress_window)
+
+
+class OpenPTCTPatientWindow(QtWidgets.QMainWindow, UIOpenPTCTPatientWindow):
+    go_next_window = QtCore.Signal(object)
+
+    def __init__(self, directory_in):
```
Tamookk, line 119: Please add docstrings to this `__init__` function.

src/Controller/TopLevelController.py
```diff
@@ -138,3 +143,8 @@ def show_image_fusion_select_window(self):
             self.image_fusion_window.update_ui()
 
         self.image_fusion_window.show()
+
+    def show_pt_ct_select_window(self):
```
Tamookk, line 147: Please add docstrings.

src/View/PTCTFusion/PETCTView.py
```diff
@@ -0,0 +1,225 @@
+from PySide6 import QtCore, QtGui, QtWidgets
```
Tamookk, line 1: Some docstrings missing in this file.

src/View/PTCTFusion/PTCTProgressWindow.py
```diff
@@ -0,0 +1,18 @@
+from PySide6 import QtCore
```
Tamookk, line 1: Missing docstrings.

test/test_pet_ct.py
```diff
@@ -0,0 +1,120 @@
...
+def test_obj(qtbot):
+    """
+    Testing Object
+    :return:
+    """
+    pet_ct = TestPETCT()
+    return pet_ct
+
+
+def test_initial_view(test_obj):
```
Tamookk, line 85: Please add docstrings for this and the 2 following functions.

Tamookk (commented):
Thank you for your explanation regarding the colours!

> I'll check with Andrew to see if he wants the patient to be the same, but im certain the image sets selected can be different.

I won't worry about this too much for now, then. It can always be changed later if needed. 

Just a few docstrings to add.

JThomas2020:
> Got one question regarding the architecture: Why do we need another class of PETCTDictContainer? From what I see when testing, the PET/CT viewer tab doesn't use anything in the PatientDictContainer, is that correct? If that's the case, why does the user need to open a non-related image before being able to open a PET/CT image? Why cant they open the PET/CT in the first open patient window? Not to mention the code repetition of having a third DictContainer class.

At Andrews request, the PET and CT images should be selected at the same time and should be initialised throught the PET/CT window. The reason there's another container for PET/CT is because the container sorts and stores the separate PET and CT data/files, which is necessary for display. Segregating the DictContainers this way does increase the amount of code, however from development we believe this is necessary as:
1. We don't want PET/CT to interact or interfere with the original image set
2. We want to store the datasets and file locations of both PET and CT, and not store it as additional info
3. Clearing just the PET/CT data from PatientDictContainer will be confusing and potentially dangerous, and require mantenance when more data is added to either the main image or the PET/CT image
4. The PET/CT container being clear is a condition that's used for checks in the program

tdlinh2712:
> > Got one question regarding the architecture: Why do we need another class of PETCTDictContainer? From what I see when testing, the PET/CT viewer tab doesn't use anything in the PatientDictContainer, is that correct? If that's the case, why does the user need to open a non-related image before being able to open a PET/CT image? Why cant they open the PET/CT in the first open patient window? Not to mention the code repetition of having a third DictContainer class.
> 
> At Andrews request, the PET and CT images should be selected at the same time and should be initialised throught the PET/CT window. The reason there's another container for PET/CT is because the container sorts and stores the separate PET and CT data/files, which is necessary for display. Segregating the DictContainers this way does increase the amount of code, however from development we believe this is necessary as:
> 
> 1. We don't want PET/CT to interact or interfere with the original image set
> 2. We want to store the datasets and file locations of both PET and CT, and not store it as additional info
> 3. Clearing just the PET/CT data from PatientDictContainer will be confusing and potentially dangerous, and require mantenance when more data is added to either the main image or the PET/CT image
> 4. The PET/CT container being clear is a condition that's used for checks in the program

Thanks for the response, I understand why an additional class might be needed now.
However, it's still concerning because:
1. The existence of `PTCTDictContainer` does not justify the existence of `OpenPTCTPatientWindow` or the amount of code repetition that it brought. I believe what it does is the same as `OpenPatientWindow` and `ImageFusionWindow`, the only difference is the condition to allow the user to proceed or not. I think I raised the same question with ImageFusionWindow before and due to some constraints, we couldn't make ImageFusionWindow class inherits OpenPatientWindow class, so this time it's probably the same problems. However, it's definitely a problem that needs to be addressed, because when our team did the hierarchy of classes (PR #168), it caused some bugs because we forgot to update the code it ImageFusionWindow. Now if there is any change in the open patient window, we would likely need to update both ImageFusionWindow and OpenPTCTPatientWindow as well and it definitely causes the program less maintainable.

2. I understand the necessity of having another class for PTCT. However, it raises a few questions: what's the point of making PatientDictContainer singleton in the first place if now there are basically three instances of it in the program? If Andrew wants another feature, such as open another type of image, or ML that requires opening another image set, do we just keep creating another clone of PatientDictContainer and repeat the code again? I don't think we have enough time to answer this question nor enough knowledge to choose the right pattern, but it's definitely an issue that needs to be addressed next year. We have created a page on the wiki for discussion about this issue (https://redmine.didymodesigns.com.au/projects/onko/wiki/OnkoDICOM's_Design_Pattern), so if you can have a look and discuss what your team's opinion would be great. Thanks!

DuyTungHa:
I also have something to add to @tdlinh2712 comments. 
I also believe that code repetition is not ideal and will cost us bugs in the future. 
Some of the abstraction methods that would resolve this issue are class inheritance, function overloading, and/or bringing repeated functions to a separate module that other places can refer to. 
Ultimately, we would have a base class/function that will be extended for specific use cases. For example, we can have a base ImageLoader.load function and other load functions from PT-CT-Fusion or Image-Fusion will extend its functionality. 
However, I understand that we may not have enough time to do this. So, it would be great if Team 19 can document this as something to do for next year.

peteqian:
> 1. The existence of `PTCTDictContainer` does not justify the existence of `OpenPTCTPatientWindow` or the amount of code repetition that it brought. I believe what it does is the same as `OpenPatientWindow` and `ImageFusionWindow`, the only difference is the condition to allow the user to proceed or not. I think I raised the same question with ImageFusionWindow before and due to some constraints, we couldn't make ImageFusionWindow class inherits OpenPatientWindow class, so this time it's probably the same problems. However, it's definitely a problem that needs to be addressed, because when our team did the hierarchy of classes (PR [Hierarchy of objects #168](https://github.com/didymo/OnkoDICOM/pull/168)), it caused some bugs because we forgot to update the code it ImageFusionWindow. Now if there is any change in the open patient window, we would likely need to update both ImageFusionWindow and OpenPTCTPatientWindow as well and it definitely causes the program less maintainable.

I'd suggest either a Factory or Abstract Factory Pattern for this situation.

JThomas2020:
> The existence of PTCTDictContainer does not justify the existence of OpenPTCTPatientWindow nor the amount of code repetition that it brought. I believe what it does is the same as OpenPatientWindow and ImageFusionWindow, the only difference is the condition to allow the user to proceed or not. I think I raised the same question with ImageFusionWindow before and due to some constraints, we couldn't make ImageFusionWindow class inherits OpenPatientWindow class, so this time it's probably the same problems. However, it's definitely a problem that needs to be addressed, because when our team did the hierarchy of classes (PR Hierarchy of objects #168), it caused some bugs because we forgot to update the code it ImageFusionWindow. Now if there is any change in the open patient window, we would likely need to update both ImageFusionWindow and OpenPTCTPatientWindow as well and it definitely causes the program less maintainable.

I understand the necessity of having another class for PTCT. However, it raises a few questions: what's the point of making PatientDictContainer singleton in the first place if now there are basically three instances of it in the program? If Andrew wants another feature, such as open another type of image, or ML that requires opening another image set, do we just keep creating another clone of PatientDictContainer and repeat the code again? I don't think we have enough time to answer this question nor enough knowledge to choose the right pattern, but it's definitely an issue that needs to be addressed next year. We have created a page on the wiki for discussion about this issue (https://redmine.didymodesigns.com.au/projects/onko/wiki/OnkoDICOM's_Design_Pattern), so if you can have a look and discuss what your team's opinion would be great. Thanks!

Thanks Linh, I do have the same concerns as well, and just to add:

1. We did attempt to use `OpenPatientWindow` as a base class for `OpenPTCTPatientWindow`, however we ran into similar issues with the design of `ImageFusionSelect`, specifically selecting/checking which leaves are checked and emitting to the loading functions. Depending on Andrew's feedback for the selection process as well (whether it needs to be from the same patient as a regular image) the conditions will need to to be changed further.
2. We have it as Singleton because multiple classes use the containers, not only that but it needs to store two image sets and all data associated, the current `PatientDictContainer` stores almost everything in a single dictionary, which is difficult to maintain if we're adding more data that is unrelated to the data already stored. Not only this but eventually when PET/CT data needs to be removed it would be difficult to segregate data for deletion. I was tempted to not use a container and store it locally on the `PETCTView`, however doing this would make accessing it impossible from the `ActionHandler` class, which means windowing functions would be excluded. PatientDictContainer will need to be refactored to allow segregation and ordering of multiple image sets however we don't have the time now to do this. I'm happy to document on redmine.

Tamookk:
I have nothing to add to the current discussion but will not approve until it has wrapped up, in case anything else is changed or added.

Thank you for adding those docstrings!

tdlinh2712:
> > The existence of PTCTDictContainer does not justify the existence of OpenPTCTPatientWindow nor the amount of code repetition that it brought. I believe what it does is the same as OpenPatientWindow and ImageFusionWindow, the only difference is the condition to allow the user to proceed or not. I think I raised the same question with ImageFusionWindow before and due to some constraints, we couldn't make ImageFusionWindow class inherits OpenPatientWindow class, so this time it's probably the same problems. However, it's definitely a problem that needs to be addressed, because when our team did the hierarchy of classes (PR Hierarchy of objects #168), it caused some bugs because we forgot to update the code it ImageFusionWindow. Now if there is any change in the open patient window, we would likely need to update both ImageFusionWindow and OpenPTCTPatientWindow as well and it definitely causes the program less maintainable.
> 
> > I understand the necessity of having another class for PTCT. However, it raises a few questions: what's the point of making PatientDictContainer singleton in the first place if now there are basically three instances of it in the program? If Andrew wants another feature, such as open another type of image, or ML that requires opening another image set, do we just keep creating another clone of PatientDictContainer and repeat the code again? I don't think we have enough time to answer this question nor enough knowledge to choose the right pattern, but it's definitely an issue that needs to be addressed next year. We have created a page on the wiki for discussion about this issue (https://redmine.didymodesigns.com.au/projects/onko/wiki/OnkoDICOM's_Design_Pattern), so if you can have a look and discuss what your team's opinion would be great. Thanks!
> 
> Thanks Linh, I do have the same concerns as well, and just to add:
> 
> 1. We did attempt to use `OpenPatientWindow` as a base class for `OpenPTCTPatientWindow`, however we ran into similar issues with the design of `ImageFusionSelect`, specifically selecting/checking which leaves are checked and emitting to the loading functions. Depending on Andrew's feedback for the selection process as well (whether it needs to be from the same patient as a regular image) the conditions will need to to be changed further.
> 2. We have it as Singleton because multiple classes use the containers, not only that but it needs to store two image sets and all data associated, the current `PatientDictContainer` stores almost everything in a single dictionary, which is difficult to maintain if we're adding more data that is unrelated to the data already stored. Not only this but eventually when PET/CT data needs to be removed it would be difficult to segregate data for deletion. I was tempted to not use a container and store it locally on the `PETCTView`, however doing this would make accessing it impossible from the `ActionHandler` class, which means windowing functions would be excluded. PatientDictContainer will need to be refactored to allow segregation and ordering of multiple image sets and ALL the aditional data generated will need to be segegated in the same way, however we don't have the time now to do this. I'm happy to document on redmine.

Thank you! I will approve once it is documented on Redmine.

JThomas2020:
> Thank you! I will approve once it is documented on Redmine.

Updated on redmine

tdlinh2712 (approved):
Team 23. Great work!

Tamookk (approved):
Team 14, good work!

PR #187, opened 2021-10-04
Bug: bytes per line for colored images

JThomas2020:
This PR fixes the bug where bytes did not account for colored pixels.


Tamookk (approved):
Quick fix, works well. Team 14.

DuyTungHa (approved):
Team23

PR #189, opened 2021-10-05
Bug: Fix Segmentation Fault

JThomas2020:
This PR adresses the issue where multiple instances of OpenPTCTPatientWindow selection can be loaded, when loading the initial data in PTCTDictContainer, it sets all other variables as none, in case the container is populated before the initial values are set.


Tamookk (approved):
Team 14 approves this. Quick but important fix!

eddieGong2604 (approved):
Team 23: Thanks for the quick fix

PR #190, opened 2021-10-05
Bug Fixes: Batch Processing 

Tamookk:
This PR fixes multiple things, mainly to do with batch processing:
- CSV files with the extension `.CSV` can now be opened in Batch CSV2ClinicalData-SR. 
- Batch ROI Name Cleaning for ROIs with standard prefixes has been fixed and now works properly. Users are given the option to enter in a new name for ROIs with standard prefixes (eg. ISO, SUV). The new name is then prefaced by the same prefix, and used to rename the ROI.
- Text has been added to the standard organ names add-on options window that directs users to a web page containing all standard organ names if they wish to add more. This URL is clickable.

### DIFF
--- a/res/stylesheet-win-linux.qss
+++ b/res/stylesheet-win-linux.qss
@@ -182,6 +182,14 @@ QLineEdit {
     font-size: 14px;
     background-color: rgb(244, 245, 245);
 }
+
+QLineEdit#BatchROICleaning {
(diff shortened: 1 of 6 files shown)


JThomas2020:
Team 19: Good work. The URL is really useful to have! Approved

PR #191, opened 2021-10-06
Fix: Select Views for Windowing

JThomas2020:
This PR adds the ability to specify the views to apply a window on:
- If DICOM view is the only active view, then windowing is automatically applied to DICOM view
- If PET/CT view or Image Fusion view is active, a popup will open with checkboxes for each active view
- The checkboxes can be selected and multiple views can be updated at the same time

This PR also moves the windowing function in src.Controller.ActionHandler to src.Model.Windowing to comply with MVC

### DIFF
--- a/src/Controller/ActionHandler.py
+++ b/src/Controller/ActionHandler.py
@@ -8,6 +8,8 @@ from src.Model.MovingDictContainer import MovingDictContainer
 from src.Model.PTCTDictContainer import PTCTDictContainer
 from src.Controller.PathHandler import resource_path
 from src.Model.ImageFusion import get_fused_window
+from src.View.WindowingUI import Windowing
+from src.Model.Windowing import windowing_model
 
 
 class ActionHandler:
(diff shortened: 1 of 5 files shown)


eddieGong2604 (changes requested):
Team 23, this PR is 7 commits behind master. Maybe you should try merge master into your branch and resolve conlicts if there are any.

eddieGong2604 (changes requested):
Team 23: Thanks for merging the master in. The code quality is OK. However, when I tested applying window into Image Fusion it produced this error:

<class 'ValueError'> too many values to unpack (expected 3) <traceback object at 0x000001CABA1DA288>
  File "C:\Users\tienl\Desktop\CSIT321\alarmancy\OnkoDICOM\src\View\WindowingUI.py", line 84, in confirmed
    windowing_model(self.text, send)
  File "C:\Users\tienl\Desktop\CSIT321\alarmancy\OnkoDICOM\src\Model\Windowing.py", line 68, in windowing_model
    get_fused_window(level, window)
ValueError: too many values to unpack (expected 3)

PR #194, opened 2021-10-08
Bug Fix: Batch Processing Unit Tests

Tamookk:
This fixes batch processing unit tests. When the batch ISO2ROI test runs, it creates an `rtss.dcm` file in the batch test data folder. This is because the RTSTRUCT in the test dataset does not strictly match the rest of this dataset when comparing UIDs.

This `rtss.dcm` file interferes with other batch unit tests, causing them to fail. This PR deletes this `rtss.dcm` file after batch ISO2ROI finishes, which fixes the other batch processing unit tests so that they pass.


DuyTungHa (approved):
Team 23 approves this PR.

JThomas2020 (approved):
Team 19: Good fix! Approved.

PR #196, opened 2021-10-13
Add: Image Fusion Addon Options

peteqian:
This PR adds an addon option for the configuration of automated image fusion. The configuration contains basic validation that is required for the image registration library.

Please note that writing values containing comma values will introduce issues with CSV therefore, another data storage format(JSON) is used.


src/Model/CalculateImages.py
```diff
@@ -88,6 +88,7 @@ def get_img(pixel_array):
     return dict_img
 
 
+
```
Tamookk, line 91: Too many blank lines, should only be two here. (PEP 8)

peteqian: I didn't see that, will make those changes.

peteqian:
> I like the change made to `View/AddOnOptions.py` - that simplifies the code quite a lot!
> 
> Are explanations of what all of these options do going to be in the user manual? If so, a message at the top saying "please consult the user manual for an explanation" or something along those lines could be added.
> 
> Just one PEP 8 compliance fix.
> 
> What is the difference between `src/View/ImageFusion/ImageFusionAddOnOption.py` and `src/View/addons/ImageFusionAddOnOption.py`? They appear to be the same file, and the former one doesn't appear to be used anywhere in the program.

As for the explanations, I cannot directly explain them as they're parameters defined in Platipy library and there is no explanations there unfortunately. This will require a further consult with the Platipy masters.

I accidentally didn't see the leftover file of the AddOnOption contained in `src/View/ImageFusion/ImageFusionAddOnOption.py` will be deleted.

PR #197, opened 2021-10-14
Change: Data Folder

Tamookk:
This PR changes what happens with the `data` folder inside OnkoDICOM. The following sequence now takes place whenever OnkoDICOM is started:
1. When the program is opened, the existence of the directory `[USER HOME]/OnkoDICOM/data` is determined. Note that this is not the same as the hidden directory, this is a separate, public directory in the user's home folder.
2. If this directory does not exist, it is created.
3. The `data` directory is then scanned, and a list of all files in this directory is created (List A). A list of all files in `[USER HOME]/OnkoDICOM/data` is also created (List B).
4. These two lists are compared. If all files in List A are in List B, then the program continues to start as normal.
5. If files in List A are missing from List B, then these files are read from, this data copied, and then the data written to a file of the same name in `[USER HOME]/OnkoDICOM/data`.

This change to the start up sequence was made to overcome permission errors when writing to these data files while running compiled and installed versions of OnkoDICOM (no OS likes it when users try to write to application directories!).

A new function has been added, `data_path`, to wrap file names for files in this data folder to ensure that the file being written to is the one in the user's home. 

Please note that when copying, original files are read from and new files are written to instead of the original files being copied because permissions carry over when you copy a file. i.e., since the original files are read-only, if they were merely copied then the new files would also be read-only.

Please test this change thoroughly on both Windows and Unix before approving.

### DIFF
--- a/src/View/batchprocessing/ROINameCleaningOptions.py
+++ b/src/View/batchprocessing/ROINameCleaningOptions.py
@@ -312,7 +312,8 @@ class ROINameCleaningOptions(QtWidgets.QWidget):
 
             # Create text entry field the ROI has a standard prefix.
             # Generate organ combobox otherwise.
-            if roi_name[0:3] in self.volume_prefixes:
+            if roi_name[0:3] in self.volume_prefixes \
+                    or roi_name[0:4] in self.volume_prefixes:
                 name_box = ROINameCleaningPrefixEntryField()
                 name_box.setEnabled(False)
             else:


Tamookk:
Please note that I have also made a small change regarding batch ROI renaming. 'PTV_' was not being picked up as a standard prefix when performing Batch ROI Name Cleaning, but this has been changed so it now is.

DuyTungHa (approved):
Team 23 approves this PR. 
However, this should just be a temporary fix because now we are storing data in multiple places: application directory, a directory in user home directory, and in the SQLite database. 
We should store all configurable data in the SQLite database. 
Please remember to write a note for future teams about this.

PR #198, opened 2021-10-19
Bug Fix: Addon Options

peteqian:
This PR is to fix issues:
1. Key and Value error when attempting to use the Study Description as the fixed and moving labels in the addon options, resulting in a crash.
2. Matrix not existing when attempting other registration methods besides "rigid".

To elaborate key point 2: a rigid matrix can be calculated when _rigid_ mode is selected. Other modes did not result in a rigid matrix with current calculations. DICOM standards allow the Spatial Registration Object to save RIGID, RIGID_SCALE and AFFINE matrices.

This PR also introduces the fix to update the fixed image and moving image labels when opening a new patient.

### DIFF
--- a/src/Controller/AddOnOptionsController.py
+++ b/src/Controller/AddOnOptionsController.py
@@ -707,3 +707,6 @@ class AddOptions:
     def show_add_on_options(self):
         self.options_window.fill_tables()
         self.options_window.show()
+
+    def update_ui(self):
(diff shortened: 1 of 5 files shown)


DuyTungHa (approved):
Team 23 approves this PR.

Tamookk (approved):
Team 14 approves this PR

PR #200, opened 2021-10-22
Bug Fix: Crash when clinical data CSV does not exist

Tamookk:
This PR fixes a bug where OnkoDICOM would crash if the clinical data CSV file path stored in the configuration database does not exist.


JThomas2020 (approved):
Team 19: Approved!

PR #208, opened 2022-08-05
Fixed float problem

Rynder1:
Fixed a problem where the min and max density values were set as floats from transect but the program itself would not work with float but required them as integers. This fixes that and allows you to select in transect and not have to change before clicking draw.


as565:
Looks good. Good logging and the code is well written, unit tests look like they are still providing good coverage from what I saw in the GitHub workflow built on this pull request.

Yc325:
Looks good.

PR #209, opened 2022-08-17
BatchProcessing - Seperate anatomy from GTV/CTV/PTV 

as565:
Changes made here are in regards to ROINameCleaningOptions.py in src/View/batchprocessing/ where the requirement was to separate anatomy changes from GTV/CTV/PTV through the GUI by either different tabs or different tables. 

There was an issue where the wrong item would be suggested when changing ROI files, this has been changed but needs to be tested with more DICOM files before I can say it is fixed. But with what I have currently, it seems to run perfectly.

The choice was made that it would be displayed as two tables within the same tab so that when the tab is ticked it will run through the whole section. So if any of the ROIs need changing or don't need changing they will be considered together and changed together. With this choice there had to be double-ups in order to display both tables together, so whenever the organ_table would be made so will the volume_table to make sure they are being initialized the same way. 

Reviews and advice are welcome.

### DIFF
--- a/src/View/batchprocessing/ROINameCleaningOptions.py
+++ b/src/View/batchprocessing/ROINameCleaningOptions.py
@@ -146,7 +146,8 @@ class ROINameCleaningOptions(QtWidgets.QWidget):
         self.volume_prefixes = []
 
         self.get_standard_names()
-        self.create_table_view()
+        self.create_table_view_organ()
+        self.create_table_view_volume()
         self.setLayout(self.main_layout)
 
     def get_standard_names(self):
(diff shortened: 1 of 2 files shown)


pedrooste (approved):
Looks good. However I didn't do any testing

src/View/batchprocessing/ROINameCleaningOptions.py
```diff
@@ -175,46 +171,87 @@ def get_standard_names(self):
...
         # Add table to the main layout
-        self.main_layout.addWidget(self.table_roi)
+        self.main_layout.addWidget(self.table_organ)
+
+    def create_table_view_volume(self):
+        """
+        Create a table to display all of the non-standard ROIs and
+        options for what to do with them.
+        """
+        # Create table
```
ryan-a-git, line 220: Needs an info log here to specify that the function has started

src/View/batchprocessing/ROINameCleaningOptions.py
```diff
@@ -175,46 +171,87 @@ def get_standard_names(self):
...
+        roi_name_header.setSectionResizeMode(2, QtWidgets.QHeaderView.Stretch)
+        roi_name_header.setSectionResizeMode(3, QtWidgets.QHeaderView.Stretch)
+
+        # Removing the ability to edit tables with immediate click
+        self.table_volume.setEditTriggers(
+            QtWidgets.QTreeView.NoEditTriggers |
+            QtWidgets.QTreeView.NoEditTriggers)
+
+        # Add table to the main layout
+        self.main_layout.addWidget(self.table_volume)
```
ryan-a-git, line 254: needs an info log here to specify the function has successfully completed

src/View/batchprocessing/ROINameCleaningOptions.py
```diff
@@ -224,8 +261,19 @@ def populate_table(self, dicom_structure, batch_directory):
                                 patients loaded.
         :param batch_directory: The directory selected for batch processing.
         """
-        # Update table column view
```
ryan-a-git, line 227: Same as previous comment for this function

src/View/batchprocessing/ROINameCleaningOptions.py
didymo: Thanks Ryan that is well pointed out.

src/View/batchprocessing/ROINameCleaningOptions.py
as565: Fixed

src/View/batchprocessing/ROINameCleaningOptions.py
as565: Fixed

src/View/batchprocessing/ROINameCleaningOptions.py
as565: Fixed

ryan-a-git (approved):

PR #211, opened 2022-09-02
FMAID2ROIName.py function and update for ROINameCleaningOptions.py

as565:
Functionality so that if the Name is changed to FMA ID it can be changed back with FMAID2ROIName.py. GUI for this function is also in this PR.

Update for ROINameCleaningOptions.py so that the volume prefix is displayed in the new name section of the table and can be edited.

I will not add logging to this PR.

edit:
PR #209 logging will be removed as per the logging meeting.

### DIFF
--- a/src/Controller/BatchProcessingController.py
+++ b/src/Controller/BatchProcessingController.py
@@ -16,6 +16,8 @@ from src.Model.batchprocessing.BatchProcessROIName2FMAID import \
     BatchProcessROIName2FMAID
 from src.Model.batchprocessing.BatchProcessROINameCleaning import \
     BatchProcessROINameCleaning
+from src.Model.batchprocessing.BatchProcessFMAID2ROIName import \
+    BatchProcessFMAID2ROIName
 from src.Model.batchprocessing.BatchProcessSUV2ROI import BatchProcessSUV2ROI
 from src.Model.DICOM.Structure.DICOMSeries import Series
 from src.Model.DICOM.Structure.DICOMImage import Image
@@ -183,6 +185,7 @@ class BatchProcessingController:
             "csv2clinicaldata-sr": self.batch_csv2clinicaldatasr_handler,
             "clinicaldata-sr2csv": self.batch_clinicaldatasr2csv_handler,
             "roiname2fmaid": self.batch_roiname2fmaid_handler,
+            "fmaid2roiname": self.batch_fmaid2roiname_handler,
         }
 
         patient_count = len(self.dicom_structure.patients)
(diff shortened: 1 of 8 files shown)


src/Model/batchprocessing/BatchProcessFMAID2ROIName.py
```diff
@@ -0,0 +1,132 @@
...
+        # Return false if RTSS has no ROIs
+        if not roi_names:
+            self.summary = "ROI_NO_FMA"
+            return False
+
+        # Convert FMA ID to ROI name 
+        rtss = self.patient_dict_container.dataset['rtss']
+        total = 0
+        progress = 40
+        step = (90 - 40)/len(roi_names)
```
nwilde, line 68: Why have 90-40 instead of just 50?

as565: Fair comment, there is no reason. Changing this now.

src/Model/batchprocessing/BatchProcessFMAID2ROIName.py
```diff
@@ -0,0 +1,132 @@
...
+                break
+
+        # Return if not found
+        if not roi_id:
+            return
+
+        # Change name of ROI to new name
+        rtss = ROI.rename_roi(rtss, roi_id, new_name)
+
+        return rtss
```
nwilde, line 132: could possibly re-write the returns to read a bit cleaner and do away with roi_id
```
for sequence in rtss.StructureSetROISequence:
    if sequence.ROIName == old_name:
        return ROI.rename_roi(rtss, sequence.ROINumber, new_name)

# Return if not found
return
```

src/Controller/BatchProcessingController.py
```diff
@@ -523,6 +526,33 @@ def batch_roiname2fmaid_handler(self, interrupt_flag,
...
+        """
+        Handles creating, starting, and processing the results of batch
+        FMA-ID2ROIName.
+        :param interrupt_flag: A threading.Event() object that tells the
+                               function to stop loading.
+        :param progress_callback: A signal that receives the current
+                                  progress of the loading.
+        :param patient: The patient to perform this process on.
+        """
+        # Get patient files and start process
```
nwilde, line 540: Some comments like these are a little bit superfluous

nwilde (commented):
Overall PR looks good, some of the comments could be removed since they don't really add much.

src/Model/batchprocessing/BatchProcessFMAID2ROIName.py
as565: That looks much better, I will be adding that change in.

src/Model/batchprocessing/BatchProcessFMAID2ROIName.py
```diff
@@ -0,0 +1,124 @@
...
+        # Get organ names and FMA IDs if they have not been populated
+        if not self.fma_ids:
+            # Get standard organ names
+            with open(data_path('organName.csv'), 'r') as f:
+                csv_input = csv.reader(f)
+                header = next(f)  # Ignore the "header" of the column
+                for row in csv_input:
+                    self.fma_ids.append(row[1])
+                    self.organ_names[row[1]] = row[0]
+                f.close()
```
nwilde, line 95: You don't need to manually close a file that is opened with a context manager (with open() as f:), it will be automatically closed once the with statement is finished.

nwilde (approved):
Like the fixes, tidies it up nicely. Nice work

src/Model/batchprocessing/BatchProcessFMAID2ROIName.py
as565: You are right, this has been fixed

nwilde (approved):

SeanOverton (approved):

PR #212, opened 2022-09-07
Batch process selecting subgroup

SeanOverton:
A new batch process that lets you select a subgroup to run all the 'checked' processes on by selecting key-values in a table populated from any clinical-data-sr files in the selected directory.


nwilde (approved):
Overall really clean code, I don't know a great deal about how the batch processing works but the code looks good and is well written.

AlexanderPadayachee:
Looks good. Docstrings are very well written. Only suggestion would be some more in line comments to describe the operation of the code.

didymo (changes requested):
Where are the lines for the logging of debugging events as required by the logging policy?

didymo (changes requested):
Where are the unit tests??

didymo (approved):

PR #213, opened 2022-09-08
Draw ROI Cursor Radius Fix #2408

nwilde:
- Allowed the drawing tool cursor to be adjusted down to a 1 pixel diameter for precise drawing
- Changed the user facing elements to display the cursor size as diameter, to better reflect the new behaviour
- Note: Back end calculations still use radius, but it is converted to diamter to display to the user


AlexanderPadayachee:
Well writen change to the code. Just a basic refactor, should be good to commit

SeanOverton (approved):

PR #216, opened 2022-09-15
Select ROI GUI refactor

pedrooste:
In order to place the draw ROI window inside of the main Onko window, we must first move the step that allows the user to select an ROI.

This will instead be placed under a button on the draw ROI window, which will then trigger the original select ROI popup.

Logging has been introduced for newly added/modified methods.

No testing is present on this branch as no changes have been fundamentally made to the select ROI functionality. One thing that could be tested is the saving restriction, where the user must select an ROI type before saving. However, this is trivial within QT due to the use of popup messages/screens.

### DIFF
--- a/src/Controller/ROIOptionsController.py
+++ b/src/Controller/ROIOptionsController.py
@@ -1,6 +1,7 @@
+import logging
+
 from src.View.ImageFusion.UITransferROIWindow import UITransferROIWindow
 from src.View.mainpage.DeleteROIWindow import *
-from src.View.mainpage.DrawROIWindow.SelectROIPopUp import SelectROIPopUp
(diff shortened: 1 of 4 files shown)


nwilde (approved):
Looks good, gave it a quick test seems to work well 👍

didymo (changes requested):
Logging should be debug not info.

nwilde (approved):

pedrooste:
Made another small change regarding the button text on the select ROI popup :)

nwilde (approved):

SeanOverton (approved):
Nice one 👍

as565 (approved):

PR #217, opened 2022-09-15
Fixed draw ROI issue when manaully drawing pixels with cursor

pedrooste:
Fixed draw ROI issue when manually drawing pixels with cursor.

This was due to the BFS source point being set as some form of float, causing the rest of the comparative array (which is used to determine if a pixel is coloured) to all be floats as well.


nwilde (approved):
Good find, gave it a quick test seems to have resolved the issue

SeanOverton (approved):
Simple but genius ;) 👍

as565 (approved):
Nice find and fix

PR #218, opened 2022-09-19
Added null check for dataset in patient_dict_container

SeanOverton:
Added a null check so the application does not crash for this BatchProcess when no SR files are found but rather behaves correctly and lets the user know that no files were found.


as565 (approved):
Looks good

nwilde (approved):
Simple fix, looks good 👍

PR #219, opened 2022-09-19
ROINameCleaning fixes

as565:
These changes are to fix the errors that Dr Miller has been facing with ROINameCleaning in the batchprocessing section.

### DIFF
--- a/src/View/batchprocessing/ROINameCleaningOptions.py
+++ b/src/View/batchprocessing/ROINameCleaningOptions.py
@@ -94,7 +94,6 @@ class ROINameCleaningDatasetComboBox(QtWidgets.QComboBox):
             self.addItem(rtss)
 
         self.setObjectName("BatchROICleaning")
-        # TODO: make changing option do nothing
 
 
 class ROINameCleaningPrefixEntryField(QtWidgets.QLineEdit):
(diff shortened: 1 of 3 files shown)


nwilde (approved):

src/View/batchprocessing/ROINameCleaningOptions.py
```diff
@@ -318,16 +319,29 @@ def populate_table(self, dicom_structure, batch_directory):
...
                 # Get the ROI name
-                roi_name = rtstruct.StructureSetROISequence[i].ROIName
+                roi_name = rtstruct.StructureSetROISequence[i].ROIName.lstrip()
 
                 # Add ROI name to the dictionary
                 if roi_name not in rois.keys():
                     rois[roi_name] = []
+                
+                # Get the suffix of the ROI name if it is a integer
+                temp = re.search(r'\d+$', roi_name)
```
SeanOverton, line 329: could use ```if roi_name.isdigit():``` here? regex might not be necessary

src/View/batchprocessing/ROINameCleaningOptions.py
```diff
@@ -318,16 +319,29 @@ def populate_table(self, dicom_structure, batch_directory):
             # Loop through each ROI in the RT Struct
             for i in range(len(rtstruct.StructureSetROISequence)):
                 # Get the ROI name
-                roi_name = rtstruct.StructureSetROISequence[i].ROIName
+                roi_name = rtstruct.StructureSetROISequence[i].ROIName.lstrip()
```
SeanOverton, line 322: .strip() rather then .lstrip() may be safer? let me know what u think

SeanOverton (approved):

src/View/batchprocessing/ROINameCleaningOptions.py
as565: I was taking out the back of the string and getting the numbers i.e. PTV1111 I would get 1111 so I could see if the length of the numbers were equal to 4. Now thinking about this I could write a couple more lines so that regex wouldn't be necessary such as roi_name[4:8].isdigit() != True and roi_name[3:7] != True respective of what prefix I'm checking in the if statement below. Interested in your thoughts

src/View/batchprocessing/ROINameCleaningOptions.py
as565: Yea it would be, I used .lstrip() in my code because one of the Dicom files had whitespace at the start of the string. It totally past my mind that it could have whitespace at the end as well. I'll put this change in and resolve this shortly.

src/View/batchprocessing/ROINameCleaningOptions.py
SeanOverton: Ah okay what you did is great then. I saw the $ sign in the regex and was assuming it was a single number you were searching eg. '1111'. Apologies, I should have thought about the fact that you had clearly written SUFFIX  in the commensts! good job

PR #221, opened 2022-09-23
Separated Fill and Draw Functionalities into Their Own Buttons #2444

Jack-Is-2048:
- Added a fill button, linked to fill functionality (practically renamed from the old draw)
- Moved draw button, linked to drawing functionality
- Fixed the bug of not being able to fill more than once

### DIFF
--- a/src/View/mainpage/DrawROIWindow/Drawing.py
+++ b/src/View/mainpage/DrawROIWindow/Drawing.py
@@ -4,6 +4,7 @@ import logging
 from PySide6 import QtCore, QtGui, QtWidgets
 from PySide6.QtGui import QColor, QPen
 from PySide6.QtWidgets import QGraphicsPixmapItem, QGraphicsEllipseItem
+from PySide6.QtCore import Slot
 
 from src.constants import DEFAULT_WINDOW_SIZE
 from src.Model.Transform import linear_transform, get_pixel_coords, \
@@ -65,6 +66,7 @@ class Drawing(QtWidgets.QGraphicsScene):
         self.max_bounds_x = self.rows
         self.max_bounds_y = self.cols
         self.fill_source = None
+        self.is_drawing = False
 
     def set_bounds(self):
         """
@@ -392,6 +394,15 @@ class Drawing(QtWidgets.QGraphicsScene):
         self.fill_source = [round(event.scenePos().x()), round(event.scenePos().y())]
         self._display_pixel_color()
 
+    @Slot(bool)
+    def set_is_drawing(self, is_drawing):
+        """
+        Function triggered when either the Draw or Fill button is pressed from the menu.
+        Sets the is_drawing bool, allows for filling or drawing from separate button presses
+        """
+        logging.debug("set_is_drawing started")
+        self.is_drawing = is_drawing
+
     def manual_draw_roi(self, event):
         """
         Draws the ROI based on the users mouse press event and position
@@ -425,7 +436,7 @@ class Drawing(QtWidgets.QGraphicsScene):
             This method is called to handle a mouse press event
             :param event: the mouse event
         """
-        if self.fill_source is None:
+        if not self.is_drawing:
             self.set_source(event)
         else:
             self.manual_draw_roi(event)
--- a/src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
+++ b/src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
@@ -3,7 +3,7 @@ import platform
 import logging
 import pydicom
 from PySide6 import QtCore, QtGui, QtWidgets
-from PySide6.QtCore import Qt, QSize, QRegularExpression
+from PySide6.QtCore import Qt, QSize, QRegularExpression, Signal
 from PySide6.QtGui import QIcon, QPixmap, QRegularExpressionValidator
 from PySide6.QtWidgets import QFormLayout, QLabel, QLineEdit, \
     QSizePolicy, QHBoxLayout, QPushButton, QWidget, \
@@ -23,6 +23,8 @@ from src.constants import INITIAL_DRAWING_TOOL_RADIUS
 
 
 class UIDrawROIWindow:
+
+    is_drawing = Signal(bool)
 
     def setup_ui(self, draw_roi_window_instance,
                  rois, dataset_rtss, signal_roi_drawn, signal_draw_roi_closed):
@@ -52,7 +54,7 @@ class UIDrawROIWindow:
         self.drawingROI = None
         self.slice_changed = False
         self.drawing_tool_radius = INITIAL_DRAWING_TOOL_RADIUS
-        self.keep_empty_pixel = False
+        self.keep_empty_pixel = True  # This constant will not change
         # pixel density
         self.target_pixel_coords_single_array = []  # 1D array
         self.draw_roi_window_instance = draw_roi_window_instance
@@ -60,6 +62,7 @@ class UIDrawROIWindow:
         self.ds = None
         self.zoom = 1.0
         self.pixel_transparency = 0.50
+        self.has_drawing = False
 
         self.upper_limit = None
         self.lower_limit = None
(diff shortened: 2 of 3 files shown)


src/View/mainpage/DrawROIWindow/Drawing.py
```diff
@@ -392,6 +394,16 @@ def set_source(self, event):
...
 
+    @Slot(bool)
+    def set_is_drawing(self, is_drawing):
+        """
+        Function triggered when either the Draw or Fill button is pressed from the menu.
+        Sets the is_drawing bool, allows for filling or drawing from separate button presses
+        """
+        logging.debug("set_is_drawing started")
+        self.is_drawing = is_drawing
+        logging.debug("set_is_drawing finished")
```
pedrooste, line 405: probably don't need a finished

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -51,14 +53,15 @@ def setup_ui(self, draw_roi_window_instance,
         self.drawingROI = None
         self.slice_changed = False
         self.drawing_tool_radius = INITIAL_DRAWING_TOOL_RADIUS
-        self.keep_empty_pixel = False
+        self.keep_empty_pixel = True  # This constant will not change
```
pedrooste, line 56: I think we need a guaranteed decision on whether this will stay in or not, otherwise, it'll end up getting lost in the code.

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -334,32 +337,19 @@ def init_layout(self):
...
-            addWidget(self.toggle_keep_empty_pixel_label)
-        self.toggle_keep_empty_pixel_box. \
-            addWidget(self.toggle_keep_empty_pixel_combo_box)
+        # Create a draw button
+        self.draw_button_row_layout = QHBoxLayout()
+        self.image_slice_number_draw_button = QPushButton()
+        self.image_slice_number_draw_button.setObjectName("ImageSliceNumberDrawButton")
+        self.image_slice_number_draw_button.clicked.connect(self.onDrawClicked)
+        self.image_slice_number_draw_button.setEnabled(False)
+        self.draw_button_row_layout.addWidget(self.image_slice_number_draw_button)
```
pedrooste, line 346: Missing a size policy

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -808,6 +792,25 @@ def onDrawClicked(self):
         """
         Function triggered when the Draw button is pressed from the menu.
         """
+        logging.debug("onDrawClicked started")
+        if hasattr(self, 'drawingROI'):
+            if self.drawingROI is not None:
+                self.is_drawing.connect(self.drawingROI.set_is_drawing)
+                self.is_drawing.emit(True)
+                self.is_drawing.disconnect(self.drawingROI.set_is_drawing)
```
pedrooste, line 800: Interesting, any reason why you have chosen this way?

test/test_view_drawing.py
```diff
@@ -95,3 +97,21 @@ def test_change_transparency_slider_value(qtbot, test_object, init_config):
...
+    test_object.draw_window.show()
+    assert test_object.draw_window.windowTitle() == "OnkoDICOM - Draw Region Of Interest"
+
+    # Assert drawing is in draw "mode"
+    test_object.draw_window.min_pixel_density_line_edit.setText("900")
+    test_object.draw_window.max_pixel_density_line_edit.setText("1000")
+    qtbot.mouseClick(test_object.draw_window.image_slice_number_fill_button, QtCore.Qt.LeftButton)
+    qtbot.mouseClick(test_object.draw_window.image_slice_number_draw_button, QtCore.Qt.LeftButton)
+    assert test_object.draw_window.keep_empty_pixel is not False
+    assert test_object.draw_window.drawingROI.is_drawing is not False
```
pedrooste, line 114: pretty sure you can simplifiy this to assert X (as it will be true)

pedrooste (changes requested):
Just a general code review, I need to read into more how the has_drawing is used.

src/View/mainpage/DrawROIWindow/Drawing.py
pedrooste: same with most, however check with ryan. My worry is that we will end up with a ton of logs.

SeanOverton: If it is actually relevent to debugging the process then leave it. But in this case the extra log on line 405 would not help in any type of debugging as the line 404 should never fail. The first log might be useful just to know when that function is being called when debugging.

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -872,12 +875,14 @@ def onDrawClicked(self):
...
                 )
                 self.slice_changed = True
+                self.has_drawing = True
                 self.dicom_view.view.setScene(self.drawingROI)
                 self.enable_cursor_diameter_change_box()
             else:
                 QMessageBox.about(self.draw_roi_window_instance,
                                   "Not Enough Data",
                                   "Not all values are specified or correct.")
+        logging.debug("onFillClicked finished")
```
SeanOverton, line 885: In this case it makes sense because of the process in between. I see that you probably added the other debug for consistency. nice job tho

SeanOverton (approved):
Dont understand this process to in depth but looks good

src/View/mainpage/DrawROIWindow/Drawing.py
Jack-Is-2048: Cheers for feedback, will resolve by removing the logging on line 405

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
Jack-Is-2048: PO, wanted to play with this feature before deciding if a "keep empty pixel" toggle is needed. This will be removed if they decide it is not needed.

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
Jack-Is-2048: This is found in both the on fill and on draw clicked functions. The reason this was done was to connect the signal to the right drawingROI object since this reference will change in between slice changes. If this were to connect once to one drawingROI object and then the user changes slice this signal would still point to that object. Therefore, it currently connects to the reference, emits the signal, and disconnects from the reference.

test/test_view_drawing.py
Jack-Is-2048: Resolved by changing to is True

as565 (approved):

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
pedrooste: Sounds good, create a ticket to follow up later

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
pedrooste: awesome, good catch too

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -403,22 +393,19 @@ def init_layout(self):
...
-        self.image_slice_number_draw_button.clicked.connect(self.onDrawClicked)
-        icon_draw = QtGui.QIcon()
-        icon_draw.addPixmap(QtGui.QPixmap(
-            resource_path('res/images/btn-icons/draw_icon.png')))
-        self.image_slice_number_draw_button.setIcon(icon_draw)
+        self.image_slice_number_fill_button.resize(
+            self.image_slice_number_fill_button.sizeHint().width(),
+            self.image_slice_number_fill_button.sizeHint().height())
+        self.image_slice_number_fill_button.clicked.connect(self.onFillClicked)
+        # TODO: Add fill icon
```
pedrooste, line 406: need a  decision on this

pedrooste (approved):
Looks good!

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
didymo: Pedro, do you mind clarifying this comment, please? I do not wish to accept the PR unless the reviewers are happy for me to do so.

pedrooste: Generally TODOs have good intention, but get left indefinitely in code. Either it needs to be removed, with the decision of not having an icon, or a story needs to be created to keep track of the todo at a later date.

This Todo is small so its okay, will have to make a story just to confirm it gets done.

PR #222, opened 2022-09-27
GUI: Back button returns to welcome window instead of application exit

djwfdev:
On both individual and batch processing screens, the back button navigates back to welcome window instead of terminating the whole application.

### DIFF
--- a/src/Model/ForceLink.py
+++ b/src/Model/ForceLink.py
@@ -1,0 +1,82 @@
+"""
+A File only used for clinical edge cases in which the RTDOSE and RTPlan
+do not reference the correct RTSTRUCT.
+This file will not force a link between IMAGE series and RTSTRUCT
+
+"""
+
+import os
+import pydicom
+import src.dicom_constants
+import logging
+
+
+def force_link(frame_overwrite, file_path, dicom_array_in):
+    """This function takes a frame of reference identifier, a file path and an
+    array of dicom ONKO DICOM widget items. This function will return 1 if the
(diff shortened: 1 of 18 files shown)


nwilde (approved):
Functionality for going back to the first page works well, don't know how to test the force link but looks good

src/Model/ForceLink.py
```diff
@@ -0,0 +1,82 @@
...
+        directory = os.listdir(file_path)
+    except FileNotFoundError:
+        logging.debug("Force link aborted, file not found")
+        return -1
+    except TypeError:
+        logging.debug("Force link aborted, file input invalid")
+        return -1
+    except NotADirectoryError:
+        logging.debug("Force link aborted, file input invalid")
+        return -1
```
pedrooste, line 47: error debug level?

pedrooste (commented):
Looks good, do we need to change the logging level?

Pretty hard to get a full idea / manually test code as its on a different fork/branch. Could be a limitation for all of our setups?

pedrooste (changes requested):
Found how to test! In pycharm there is a PR tab you can access from the main Onko Project. Thanks @nwilde 

There is a problem when spamming between the "Individual Patient Curation" and main welcome page. Eventually leads to a segmentation fault, which can differ on your end. This was probably not originally introduced in this PR, but should get looked at?

Not completely sure what the Force Link does, but when I select nothing it seems to not do anything. I would suggest it has an initial disabled state if this is the case

PR #227, opened 2022-10-05
Force link code, Back button implementation

as565:
"On both individual and batch processing screens, the back button navigates back to the welcome window instead of terminating the whole application.

Force link button added to the single patient window. Please be aware that the force link button will overwrite Dicom files in the directory." - Alexander Padayachee

Edit: I removed loguru and replaced it with import logging, as this started causing issues for our repo and tests were failing. I did a search through the whole system on VSCode so hopefully, this will not arise again.


nwilde (changes requested):
Code looks good, I don't know how to test the force link functionality so I assume you guys have tested that rigorously enough. However, I'm still encountering the bug where if you navigate back and forth between individual patient curation and the welcome screen a few times the program crashes. I'd make sure that when a user goes back to the welcome screen it completely resets/kills the instance of the program that was created. It's not that you would expect a user to do this, but if there is something being left behind that is memory or thread unsafe from a user just doing it once, then it could cause weird bugs for other sections of the program.
image

djwfdev:
> 

Cannot seem to replicate this issue of crashing with some code changes but is the "Multithreading with maximum threads" an issue?

nwilde:
> @nwilde Cannot seem to replicate this issue of crashing with some code changes but is the "Multithreading with maximum threads" now shows up only once, is that still an issue?

Yeah that seems to have fixed it, nice work 👍

PR #229, opened 2022-10-07
Kaplan meier

Santi2468:
Created Keplan Meier Plot tab within batch processing to generate plot from user selected columns


requirements.txt
```diff
@@ -18,3 +18,5 @@ matplotlib
 shapely==1.8a3
 python-Levenshtein
 opencv-python<=4.5.3.56
+kaplanmeier
+matplotlib
```
pedrooste, line 22: matplotlib is already in this file, has Ashley confirmed on kaplanmeir

src/Controller/BatchProcessingController.py
```diff
@@ -703,3 +708,46 @@ def create_timestamp(cls):
...
         return time_stamp
+
+    def set_kaplanmeier_targetCol(self, target):
+        self.kaplanmeier_targetCol = target
+
+    def set_kaplanmeier_DurationOfLifeCol(self, duration):
+        self.kaplanmeier_DurationOfLifeCol = duration
+
+    def set_kaplanmeier_AliveOrDeadCol(self, AliveOrDead):
+        self.kaplanmeier_AliveOrDeadCol = AliveOrDead
```
pedrooste, line 719: pretty sure this against python convention as we can directly access?

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
...
+
+class BatchProcessKaplanMeier(BatchProcess):
+    """
+    This class handles batch processing for the Clinical Data 2 CSV
+    process. Inherits from the BatchProcessing class.
+    """
+    # Allowed classes for ClinicalDataSR2CSV
+    allowed_classes = {
+        # Comprehensive SR
+        "1.2.840.10008.5.1.4.1.1.88.33": {
```
pedrooste, line 19: theres a generic file for these ID's would be good to chuck it in there. I cant remember where it is but shouldn't be hard to find

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
...
+    def start(self):
+        """
+        Goes through the steps of the ClinicalData-SR2CSV conversion.
+        :return: True if successful, False if not.
+        """
+        # Stop loading
+        if self.interrupt_flag.is_set():
+            self.patient_dict_container.clear()
+            self.summary = "INTERRUPT"
+            return False
```
pedrooste, line 61: extract this out into a separate method "check_interrupt" or similar. call as need

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
...
+        Searches the patient dict container for any SR files containing
+        clinical data. Returns the first SR with clinical data found.
+        :return: ds, SR dataset containing clinical data, or None if
+                 nothing found.
+        """
+        datasets = self.patient_dict_container.dataset
+
+        for ds in datasets:
+            # Check for SR files
+            if datasets[ds].SOPClassUID == '1.2.840.10008.5.1.4.1.1.88.33':
```
pedrooste, line 107: find the file (refer above comment)

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
...
+
+        return data_dict
+
+    def plot(self,data_dict, target, DurationOfLife, AliveOfDead):
+        """
+            Creates plot based off input columns
+        """
+        logging.debug("Plot function called")
+        try:
+            logging.debug("Creating/Displaying plot")
```
pedrooste, line 141: Not much need for the dual logs

src/Controller/BatchProcessingController.py
pedrooste: ah, actually I see that this file uses setters throughout. It would be better then to follow what was previously done as you have :)

src/View/batchprocessing/KaplanMeierOptions.py
```diff
@@ -0,0 +1,87 @@
...
+        self.comboboxDuration.setStyleSheet(self.stylesheet)
+        DurationLabel = QtWidgets.QLabel(
+            "Duration of Life Column")
+        DurationLabel.setStyleSheet(self.stylesheet)
+        self.duration_layout.addWidget(DurationLabel)
+        self.duration_layout.addRow(self.comboboxDuration)
+
+        # alive or dead widgets
+        self.comboboxAliveOrDead = QComboBox()
+        self.comboboxAliveOrDead.setStyleSheet(self.stylesheet)
```
pedrooste, line 58: You seem to switch between camel and snake case naming conventions. Any reason? Otherwise pick one (snake pref)

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
+import kaplanmeier as km
+import matplotlib as mp1
+mp1.use("TkAgg")
```
pedrooste, line 3: Tkinter isnt included as a dependency? This would cause the PR not to build if it isn't installed

src/View/batchprocessing/KaplanMeierOptions.py
```diff
@@ -0,0 +1,87 @@
...
+        AliveOrDeadLabel = QtWidgets.QLabel(
+            "Alive or Dead Column")
```
pedrooste, line 60: Most other labels are run through a _translate function. Have a look into it on the other pages as I think it's needed for accessibility.

pedrooste (changes requested):
see comments

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
...
+        Reads clinical data from the found SR file.
+        :param sr_cd: the clinical data SR dataset.
+        :return: dictionary of clinical data, where keys are attributes
+                 and values are data.
+        """
+        data = sr_cd.ContentSequence[0].TextValue
+
+        data_dict = {}
+
+        data_list = data.split("\n")
```
nwilde, line 125: Don't know what the input data looks like but maybe consider using `splitlines()` in case you get data with `\r\n`

src/Model/batchprocessing/BatchProcessKaplanMeier.py
```diff
@@ -0,0 +1,163 @@
...
+        for row in range(len(data_list)):
```
nwilde, line 126: This should probably just be iteration over the `data_list` instead of using range. `for row in data_list:` or something similar

nwilde (commented):
Overall looks good apart from a few tweaks. 
Other question is why has the `0.35` file and the vtk wheel been added to the commit? I assume it was just an accident, probably just need to remove them

didymo (changes requested):
deleted

PR #230, opened 2022-10-08
Draw ROI GUI and Slider fix

shirley9625:
Add Zoom slider and Cursor Diameter slider, fix Transparency slider, add fill icon and change buttons layout.


res/stylesheet.qss
```diff
@@ -75,10 +75,16 @@ QPushButton#SUV2ROI_Button {
 }
```
pedrooste, line 75: I know most of your dev is on MAC, do we need someone to test it on windows to double-check this stylesheet?

res/stylesheet-win-linux.qss
```diff
@@ -107,10 +107,16 @@ QPushButton#SUV2ROI_Button {
...
 
 QPushButton:disabled, QPushButton[QPushButtonClass="success-button"]:disabled,
+QPushButton[QPushButtonClass="draw-roi-button"]:disabled,
+QPushButton[QPushButtonClass="reset-button"]:disabled,
 QPushButton[QPushButtonClass="fail-button"]:disabled {
     background-color: #b3b3b3;
 }
 
+QPushButton[QPushButtonClass="draw-roi-button"] {
+    min-height: 30px;
```
pedrooste, line 117: did we get feedback if we could use the 30px? I think it looks good though

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -330,54 +362,24 @@ def init_layout(self):
         self.transparency_slider.setTickInterval(10)
         self.transparency_slider.setValue(50)
         self.transparency_slider.setObjectName("TransparencySlider")
-        self.transparency_slider.resize(self.transparency_slider.sizeHint().width(),
-                                        self.transparency_slider.sizeHint().height())
+        self.transparency_slider.setSizePolicy(QSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed))
```
pedrooste, line 365: The transparency slider also needs to be disabled until a ROI is drawn

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -714,6 +774,19 @@ def onZoomOutClicked(self):
             setText("{:.2f}".format(self.dicom_view.zoom * 100) + "%")
         self.draw_roi_window_viewport_zoom_input.setCursorPosition(0)
 
+    def zoom_slider_value_changed(self):
```
pedrooste, line 777: The zoom slider also needs to change when the top zoom in/out toolbar buttons are pressed. You could probably do this using a signal and slot, otherwise, check how the slices have done it.

This could be a problem as It would call this method, you could always pass an extra boolean to determine whether to carry out the updated logic.

pedrooste (changes requested):
Sliders look so much better :)

res/stylesheet.qss
shirley9625: Don't worry about this, I wrote everything on a windows desktop(I major use windows), and tested everything. Fix the mac stylesheet issue just a  small bug I found when I tested on my MacBook. Just need to get someone to test it on Linux or I can test it after I finish work so should be fine.

res/stylesheet-win-linux.qss
shirley9625: Still waiting for Helena's response. But, It only applies the 30px on the Draw ROI interface buttons. Does not affect other buttons.

didymo: I just hassled Helena.

shirley9625: I had a deep investigation on the stylesheet yesterday, It's not just about the QPushbutton size, the QLineEdit min-height is 36px as well. I think this is one of the reasons. If we gonna change the QPushbutton min-height straight away, we have to change the QLineEdit as well(it's gonna involve more stuff). So the safe and easy way is only to apply these changes on Draw ROI UI.

as565 (approved):
Looks good

res/stylesheet.qss
pedrooste: Im on Ubuntu :), worked for me so all good then!

PR #231, opened 2022-10-10
Added Automated Hole-Filling Functionality When Drawing ROIs #2457

Jack-Is-2048:
- Added automated hole-filling functionality when drawing ROIs.
  - Functionality works for both 2D and 3D fills.
  - When drawing to have holes be automatically filled enter an "internal hole size" greater than 0.
  - To not have holes be automatically filled enter 0 for an "internal hole size" or leave the field blank.

- Also contains numerous fixes for a rebase error that occurred while adding 3D fill functionality.


as565 (approved):
Looks good, nice work!

PR #233, opened 2022-10-13
Add clear button and change reset button

shirley9625:


src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -1132,6 +1130,45 @@ def onBoxDrawClicked(self):
         self.disable_cursor_diameter_transparency()
         self.has_drawing = False
 
+    def onClearClicked(self):
```
SeanOverton, line 1133: I would recommend running flake8 on this file to meet the styling requirements. this can be done in the cmd prompt by running ```flake8 src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py```. As a specific example, the function name should be snake_case: on_clear_clicked

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
```diff
@@ -1132,6 +1130,45 @@ def onBoxDrawClicked(self):
...
+        self.dicom_view.update_view()
+        self.isthmus_width_max_line_edit.setText("5")
+        self.internal_hole_max_line_edit.setText("9")
+        self.min_pixel_density_line_edit.setText("")
+        self.max_pixel_density_line_edit.setText("")
+        if hasattr(self, 'bounds_box_draw'):
+            delattr(self, 'bounds_box_draw')
+        if hasattr(self, 'drawingROI'):
+            delattr(self, 'drawingROI')
+            self.has_drawing = False
```
SeanOverton, line 1147: as this line is the only difference in the two functions: **onClearClicked** and **onResetClicked**, you could possibly reduce the repetitive code here. For example, you could just call the reset function from the clear function

SeanOverton (commented):
Looks good, maybe just review the couple comments I left, flake8 is the main one for styling reasons to fit into the rest of the propject.

src/View/mainpage/DrawROIWindow/UIDrawROIWindow.py
shirley9625: To be honest I tried before, but this will come up: qt.core.qmetaobject.connectslotsbyname: QMetaObject::connectSlotsByName: No matching signal for on_clear_clicked(). I have to follow the signature style to be used with connectSlotsByName instead of snake_case. More info here: https://forum.qt.io/topic/32092/no-matching-signal-for-solved

sjswerdloff: Qt is known to have stuck with CamelCase.
Its one of those things that one has to live with...

SeanOverton: Ah okay i see, makes sense then

as565 (approved):
Looks good, nice changes

Yc325 (approved):

PR #235, opened 2022-10-13
Added 3D Fill Progress Window

pedrooste:
**Problem/Motivation**
When a user is using the 3d fill functionality, it displays no feedback to the user. This is not desirable and can cause the user to think the program has crashed

**Proposed Resolution**
Once the 3d fill has started processing, display some sort of processing feedback to the user (loading bar, etc)


as565 (approved):

ryan-a-git (approved):
Excellent

Yc325 (approved):

PR #238, opened 2022-10-17
Machine learning testing pipeline

SeanOverton:
The testing pipeline for the Machine Learning training capability that has been introduced.

To summarise; this stage allows for the actual use of the previously generated model from the Batch Process which means values can be predicted in the clinical data of a patient.


src/Model/MachineLearningTester.py
```diff
@@ -0,0 +1,181 @@
...
+
+        return True
+
+    def save_into_csv(self, path_to_save):
+        """
+        Saves predicted values into csv.
+        Parameters:
+            str : path_to_save : location to save csv
+        """
+        # TODO: is all this necessary?
```
pedrooste, line 104: TODO

src/View/mainpage/MLTab.py
```diff
@@ -0,0 +1,656 @@
...
+
+        # once run will trigger results popup window
+        ml_results_window = MLResultsWindow()
+        ml_results_window.set_results_values(results)
+        ml_results_window.set_ml_tester(ml_tester)
+
+        ml_results_window.exec_()
+
+
+class MLResultsWindow(QtWidgets.QDialog):
```
pedrooste, line 517: good to have a separate file per class

pedrooste (commented):
Small comments but overall the code looks good.

Not sure how to test from my end

src/View/mainpage/MLTab.py
SeanOverton: True

PR #240, opened 2022-10-18
Alpha value persistance fix

nwilde:
- Added alpha value to the AddOn options
- Alpha value now saved in a data_path config file for DrawROI
- DrawROI window now fetches and saves alpha value to the config file when opening and closing
- Changed the alpha value UI element from a plain textbox to a spin box, since the text box was not properly validating the input values. Spinbox fixes this issue by fully bounding the input to the desired parameters, though either using the controls or typing


nwilde:
Need to confirm with Dr Miller what the minimum, maximum and default values should be for alpha value. Current placeholders are 0.2-0.9 with default of 0.5

SeanOverton (approved):

SeanOverton:
> Need to confirm with Dr Miller what the minimum, maximum and default values should be for alpha value. Current placeholders are 0.2-0.9 with default of 0.5

Sounds good, just a reminder to make sure it cannot be set to zero (0) as I saw this value is divided in a few places. Currently you have that minimum limit though as you state.

PR #241, opened 2022-10-20
Batch processing tabs styling

SeanOverton:
Changing the styles of the batch processing tabs to be on the left side instead of the top of the window for easier navigation. 

This was more annoying then it had to be... feel free to have a look and make any recommendations as the styling. Some of the styling was broken by the changes so a basic styling was added. Open to any suggestions and feedback.

For reference:
Before changes:
image

After the changes:
image


nwilde (changes requested):
When opening the main window your style sheet changes are affecting all qt elements of that type, you'll need to use ID selectors and object names to only affect the batch processing stuff. https://doc.qt.io/QT-5/stylesheet-syntax.html
image

Back in the batch processing window, not a huge fan of how the tabs are laying out the label/checkbox, is it possible to align them in a row? would look cleaner

SeanOverton:
> When opening the main window your style sheet changes are affecting all qt elements of that type, you'll need to use ID selectors and object names to only affect the batch processing stuff. https://doc.qt.io/QT-5/stylesheet-syntax.html image
> 
> Back in the batch processing window, not a huge fan of how the tabs are laying out the label/checkbox, is it possible to align them in a row? would look a bit cleaner image

To your first point, definitely agree, thought that could happen, but forgot to check anywhere. so i will sort that out.
For moving the checkbox to the left or right of the text... sounds simple... but proved to be very difficult (spent hours playing around with this).

nwilde (approved):
Thanks for fixing the main window. 
Subjectively, I wouldn't merge this in until a way to lay out the tabs properly is found. But if Dr Miller is happy with it then it's ultimately up to him

SeanOverton:
> Thanks for fixing the main window. Subjectively, I wouldn't merge this in until a way to lay out the tabs properly is found. But if Dr Miller is happy with it then it's ultimately up to him

Agreed.

nwilde:
Only solution I can think of is to place the checkbox and label (instead of using the built-in tab title) inside a QWidget that you insert into the tab, lay the check and label out as a row, then once the rotation shenanigans are done it would preserve the original widget in row form. But I assume that's not straight forward to do

as565 (approved):
Looks good, I had the same comments about the positioning of the tabs but as nick said if dr miller is happy, all good. Nice work! GUIs are not fun.

PR #251, opened 2022-11-20
Fixes and workarounds needed to get 0.05.02-beta running on Ubuntu 20.04

sjswerdloff:
Workaround: wrapping import of pymedphys with try/except due to recursive import that fails
Fix: eliminated casting of bitmask to int before applying bitwise OR 
Fix: wrapping initialisation of QSpacerItem with try/except to address signature variations

I'll investigate why the import of pymedphys is recursing in to areas of pymedphys that are not needed  for pseudonymisation, but for now the workaround will give the user an error when they try to pseudonymise rather than prevent them from launching OnkoDICOM at all,.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "onkodicom"
-version = "0.1.0"
+version = "0.5.2"
 description = ""
 authors = ["Ashley Maher <ashley.maher@didymodesigns.com.au>"]
 license = "LGPL"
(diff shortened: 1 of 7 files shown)


didymo:
CHekcs failing. Could this affect the executable given the condtional loads and the executable needs everything bundled?.

sjswerdloff:
The failures don't point to anything in the changes I made.
It seems like a lot of the data "expected to be present" is not?
I can create another branch and pull request that has just an additional blank line to see, or we could try to kick off another test on the main branch

sjswerdloff:
I'll work on the tests in my development environment (Ubuntu 22.04)

sjswerdloff:
Updated PR to point out that this is for Ubuntu 20.04.
I started with 22.04, but plastimatch won't install as a package on that.
I also included a fix for the slice location issue and updated the "product version" in pyproject.toml

sjswerdloff:
dropping back to python 3.9 addressed a lot of problems...
But these changes will make resolving the pymedphys related import issue connected to Python 3.10 easier to resolve

PR #265, opened 2023-02-15
Fix: Added error handling when StudyID is not present

peteqian:


src/View/ImageLoader.py
```diff
@@ -166,8 +166,11 @@ def load(self, interrupt_flag, progress_callback):
 
                     return True
         else:
-            self.load_temp_rtss(path, progress_callback, interrupt_flag)
-
+            try:
+                self.load_temp_rtss(path, progress_callback, interrupt_flag)
+            except ValueError as e:
+                raise ImageLoading.NotAllowedClassError(e)
```
sjswerdloff, line 172: is the ValueError from load_temp_rtss due to an object not being in the allowed list of SOP Classes?
If so, great.
If the error might be something else... then that's not the appropriate error class to use.

src/Model/ROI.py
```diff
@@ -953,6 +955,9 @@ def create_initial_rtss_from_ct(img_ds: pydicom.dataset.Dataset,
         raise ValueError(
             "The given dataset is missing a required tag 'StudyInstanceUID'")
 
+    if hasattr(rt_ss, "StudyID") is False or rt_ss.StudyID == "":
+        raise ValueError("The given dataset is missing a required tag 'StudyID'")
```
sjswerdloff, line 959: technically, it's missing a tag that is required by OnkoDICOM.
The DICOM standard identifies StudyID as type 2 (attribute must be encoded, but an empty value is allowed), not type 1 (attribute must be encoded with a value).

src/View/OpenPatientWindow.py
```diff
@@ -546,16 +546,17 @@ def on_loading_error(self, exception):
...
-            self.progress_window.close()
+            errorMessage = "Selected files cannot be opened as they are not a DICOM-RT set."
         elif type(exception[1]) == ImageLoading.NotAllowedClassError:
-            QMessageBox.about(self.progress_window, "Unable to open selection",
-                              "Selected files cannot be opened as they contain"
-                              " unsupported DICOM classes.")
-            self.progress_window.close()
+            errorMessage = "Selected files cannot be opened as they contain unsupported DICOM classes."
+        else:
+            errorMessage = "An error occurred while loading the selected files. Please try again."
```
sjswerdloff, line 556: a user should only be encouraged to try again if this was a temporary error (and retrying might succeed).
The user might want to investigate the error and if they can correct it, try again.
If the data are missing StudyID... there might not be a fix (unless they want to do surgery on all of the DICOM objects missing the StudyID).
What is the intent/expectation here?

src/Model/ROI.py
peteqian: So this shouldn't throw an error.

src/View/ImageLoader.py
peteqian: There is an error thrown (besides the newly created one) that occurs in this workflow. 

It also occurs in `src/Model/ROI.py` when StudyInstanceUID is not present.

src/View/OpenPatientWindow.py
peteqian: What I've done is added a generic error handling here as previously there was no generic error handling. So if there was an uncaught error, Line: 556 would be displayed. This could definitely be refined further as I'm not entirely sure what you would like the application to behave if there's an uncaught exception.

Given if the StudyID is missing, this would've thrown `NotAllowedClassError` and Line:554 would've been displayed but I'm also not entirely sure if this is the correct intention and would further clarification.

sjswerdloff:
See #305 (took the basic idea from here)

PR #268, opened 2023-02-27
Lock the shapely version

celeron533:
shapely 2.0.0 changed the API which causes test fail.

This issue first appeared in:
https://github.com/didymo/OnkoDICOM/actions/runs/3955459739


celeron533:
Migrating to Shapely 1.8 / 2.0
https://shapely.readthedocs.io/en/stable/migration.html

pyproject.toml
```diff
@@ -30,7 +30,7 @@ pytest-qt = { version = ">=4.0.2"}  # groups = ["user", "dev"]
 pytest = { version = "*"}  # groups = ["user", "dev"]
 pyside6 = { version = "6.3.0"}  # groups = ["user", "dev"]
 vtk = { version = "*"}  # groups = ["user", "dev"] 
-shapely = ">=1.8.1"  # groups = ["user", "dev"]
+shapely = ">=1.8.1,<=1.8.5"  # groups = ["user", "dev"]
```
sjswerdloff, line 33: <2.0

There might be more bug fixes to 1.8

requirements.txt
```diff
@@ -15,7 +15,7 @@ pyradiomics
 pyside6==6.2.0
 vtk
 matplotlib
-shapely>=1.8.1
+shapely>=1.8.1,<=1.8.5
```
sjswerdloff, line 18: <2.0

Might be more bug fixes to 1.8
semver approach suggest that a major version change is the interface break point

sjswerdloff (changes requested):
Change upper bound to <2.0

sjswerdloff:
Sklearn vs scikit-learn was a build failure
Is shapely dependent on sklearn? 
Did  shapely also change its dependency in 2.0 to scikit-learn

sjswerdloff (commented):
The bug fix value should be incremented on each PR

requirements.txt
celeron533: build pass and test pass when `shapely>=1.8.1,<2.0`

celeron533:
> Sklearn vs scikit-learn was a build failure Is shapely dependent on sklearn? Did shapely also change its dependency in 2.0 to scikit-learn

the output from my repo's github action looks good as the following
```
Collecting sklearn
  Downloading sklearn-0.0.post1.tar.gz (3.6 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting imblearn
```
while here it is shown as below even they are in the same version
```
Collecting sklearn
  Downloading sklearn-0.0.post1.tar.gz (3.6 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [18 lines of output]
      The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.
      
      Here is how to fix this error in the main use cases:
      - use 'pip install scikit-learn' rather than 'pip install sklearn'
      - replace 'sklearn' by 'scikit-learn' in your pip requirements files
        (requirements.txt, setup.py, setup.cfg, Pipfile, etc ...)
      - if the 'sklearn' package is used by one of your dependencies,
        it would be great if you take some time to track which package uses
        'sklearn' instead of 'scikit-learn' and report it to their issue tracker
      - as a last resort, set the environment variable
        SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True to avoid this error
      
      More information is available at
      https://github.com/scikit-learn/sklearn-pypi-package
      
      If the previous advice does not cover your use case, feel free to report it at
      https://github.com/scikit-learn/sklearn-pypi-package/issues/new
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
```

celeron533:
@sjswerdloff updated the dependency of scikit-learn in requirement.txt.
The item in pyproject.toml is already up-to-date.

PR #269, opened 2023-03-04
Cache test file

celeron533:
Cache the test file (DICOM-RT-02.zip). This saves at least 3 minutes (sometimes 5 minutes) for each build.

Note: pip is not cached if we always want to use the **latest** version of packages

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -14,9 +14,10 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Set up Python 3.9
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: 3.9
+        # cache: 'pip'
 
     - name: Install plastimatch
       run: |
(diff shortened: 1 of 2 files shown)


didymo:
This PR appears to be using python3.9, is this interpretation correct? Is OnkoDICOM not using python3.10 at the moment?

celeron533:
> This PR appears to be using python3.9, is this interpretation correct? Is OnkoDICOM not using python3.10 at the moment?

Hu Ashley,
As far as I know, both the wiki page and github action are using Python 3.9
https://github.com/didymo/OnkoDICOM/wiki/Installation-Instructions

Running on Python 3.10 need some adjustments, especially lock the dependicies in requirements.txt

celeron533:
I can  create a build matrix to build and test on both Python 3.9 and 3.10

celeron533:
> I can create a build matrix to build and test on both Python 3.9 and 3.10

of course, in another PR

sjswerdloff (approved):
LGTM

sjswerdloff:
> > I can create a build matrix to build and test on both Python 3.9 and 3.10
> 
> of course, in another PR

That would be great.
But make the 3.10 optional in terms of passing.
We may have problems being compatible with both 3.9 and 3.10 and I want to start with what we know should work.  We will probably have to pin packages in the 3.10 build and I would like to try to have both 3.9 and 3.10

PR #271, opened 2023-03-20
Build and test on Windows

celeron533:
However, pytest with certain UI related cases are not supported/stable on Windows 2022, especially python 3.10.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -3,16 +3,48 @@ name: Python application
 on: [push, pull_request]
 
 jobs:
-  build:
+  fetch_cache:
+    name: fetch cache
+    runs-on: 'ubuntu-latest'
+    env:
+      TestDataUrl: 'https://zenodo.org/record/5147737/files/DICOM-RT-02.zip'
+      
+    steps:
+    - name: Cache Test Data
+      id: cache-TestData
+      uses: actions/cache@v3
+      with:
+        enableCrossOsArchive: true
+        lookup-only: true
+        path: DICOM-RT-02.zip
+        key: ${{ env.TestDataUrl }}
+
+    - name: Download Test Data
+      if: steps.cache-TestData.outputs.cache-hit != 'true'
+      run: |
+        sudo apt-get install wget
+        wget -O DICOM-RT-02.zip $TestDataUrl
+
+
+  build_Linux:
+    needs: fetch_cache
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.9', '3.10']
-    runs-on: ubuntu-20.04
-    timeout-minutes: 60
+        python:
+          - '3.9'
+          - '3.10'
+        os:
+          - 'ubuntu-20.04'
+          # - 'ubuntu-22.04'
+
+    name: Build - ${{ matrix.os }} - Python ${{ matrix.python }}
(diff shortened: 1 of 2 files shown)


sjswerdloff (approved):

.github/workflows/pythonapp.yml
```diff
@@ -25,14 +57,16 @@ jobs:
...
     - name: Install plastimatch
       run: |
-       sudo apt-get install plastimatch
+        sudo apt-get update
+        sudo apt-get install plastimatch
 
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip wheel setuptools
         pip install -r pre-requirements.txt
```
sjswerdloff: shouldn't we be using poetry to do the installations?

celeron533: Poetry has *lot* of problems. I'll create an Issue to discuss.

didymo: The intention has been to use poetry; however, sorting dependencies has been a blocker.

sjswerdloff: In the Windows + Python 3.10 test:
`platform win32 -- Python 3.10.10, pytest-7.2.2, pluggy-1.0.0 -- C:\hostedtoolcache\windows\Python\3.10.10\x64\python.exe`

is win32 indicating a 32 bit platform or just a legacy name for MS Windows?

celeron533: > The intention has been to use poetry; however, sorting dependencies has been a blocker.

There are some issues on Poetry:
1. It cannot install `pyradiomics` (which is not following pep517)
2. The `*requirements.txt` files are out-of-sync with `pyproject.toml`
3. `poetry.lock` is not tracked in the repo
4. Poetry is slow (slightly)
5. From my experiments, I manually sync the `pyproject.toml` with workable dependencies from `*requirements.txt`, after a workaround to install `pyradiomics`, it fails more test cases than traditional `requirements.txt`

celeron533: And personally, I prefer the repo to have the install, build, test instruction with the source code (markdown) rather than wiki page. Otherwise, the wiki page and the software itself will out-of-sync.
Wiki page is better for a brief instruction, trouble shooting, or release notes, etc.

celeron533: PS.
The previous github action yaml file was using requirements.txt while the wiki page is using poetry (but no poetry.lock file)

celeron533: > In the Windows + Python 3.10 test: `platform win32 -- Python 3.10.10, pytest-7.2.2, pluggy-1.0.0 -- C:\hostedtoolcache\windows\Python\3.10.10\x64\python.exe`
> 
> is win32 indicating a 32 bit platform or just a legacy name for MS Windows?

Legacy name.
It shows win32 on my x64 python+windows

celeron533: Hi @didymo , after this PR, I'll work on Poetry with workarounds and experimentals.

didymo: G'day @celeron533 , the poetry branch has been updated for your use. The pyproject.toml needs an update which I am attempting to source.

celeron533:
Let's working on the things one-by-one.
Windows build first, then with Poetry. 🥇

PR #272, opened 2023-03-20
Poetry

didymo:


celeron533:
Since the GitHub Actions for this PR is still based on traditional pip install -r `requiremants.txt`, I may create a new PR with modified pipeline which is using Poetry to install the dependencies.

celeron533:
Should we also check-in the `poetry.lock` file?
https://stackoverflow.com/questions/61037557/should-i-commit-lock-file-changes-separately-what-should-i-write-for-the-commi/74045098#74045098

PS. In NodeJS, developers always check-in `package-lock.json` to scm.

didymo:
> Since the GitHub Actions for this PR is still based on traditional pip install -r `requiremants.txt`, I may create a new PR with modified pipeline which is using Poetry to install the dependencies.

yes, please.
The intention always was to do so, but poetry or, more precisely, dependencies hell has put this off repeatedly.  At the moment numpy version is being pinned to 1.22, but that still fails, it needs to be discovered as to why. Pyside6 needs to increase to 6.4 but is pinned at 6.3 at the moment. Too many dependencies.

didymo:
> Should we also check-in the `poetry.lock` file? https://stackoverflow.com/questions/61037557/should-i-commit-lock-file-changes-separately-what-should-i-write-for-the-commi/74045098#74045098
> 
> PS. In NodeJS, developers always check-in `package-lock.json` to scm.

No.

The poetry lock file was causing issues during the installation so it was removed.

PR #280, opened 2023-08-11
Update version of pymedphys, eliminate pyradiomics from pyproject and specify post install of pyradiomics==3.0.1

sjswerdloff:
just testing


sjswerdloff:
@Vii17b @AnnieAnNguyen @danielHorton4001 @PhilThom
I forgot to update the poetry.lock file 
i.e.
`poetry update`

sjswerdloff:
While this is a relatively small change, I think it may work out some of the issues we were having with pymedphys and python 3.10

PR #282, opened 2023-08-11
updated pymedphys version

PhilThom:
testing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ scikit-image = "*"
 fuzzywuzzy = "*"
 PyWavelets = "1.3.0"
 networkx = "*"
-pymedphys = ">=0.35,<=0.37.1"
+pymedphys = ">=0.35,<=0.37.2"
 # until 20230611, the pyradiomics does not have wheel for py3.10. Will install it with pip later.
 pyradiomics = {version = "*", python = "<3.10"}
 pyside6 = "6.3.0"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ scikit-image = "*"
 fuzzywuzzy = "*"
 PyWavelets = "1.3.0"
 networkx = "*"
-pymedphys = ">=0.35,<=0.37.1"
+pymedphys = ">=0.35,<=0.37.2"
```
sjswerdloff, line 31: This will break.  0.37.2 and later had the pseudonymisation interface disabled due to it's susceptibility to rainbow table attack.  0.39 had that fixed by adding pepper to the hashing.

sjswerdloff (changes requested):
need different version constraint

pyproject.toml
PhilThom: I changed the version to 0.39.3, my python version is 3.10.11

sjswerdloff (approved):

pyproject.toml
```diff
@@ -28,7 +28,7 @@ scikit-image = "*"
 fuzzywuzzy = "*"
 PyWavelets = "1.3.0"
 networkx = "*"
-pymedphys = ">=0.35,<=0.37.1"
+pymedphys = ">=0.35,<=0.39.3"
```
sjswerdloff, line 31: whoops, no.  This allows any of the pymedphys versions between 0.36 and 0.39.3, and it should only allow 0.39.3 and later.  I specified it so that it would use 0.39.3 up to but not including 0.40, if you want to specify it to just be >= 0.39.3

sjswerdloff (changes requested):

PR #285, opened 2023-08-11
Updated version number of pymedphys

Vii17b:


sjswerdloff:
LGTM. (Looks Good To Me).
See my comment over at #280 (I forgot to update the poetry.lock file).
Were I to merge this in, I would do a squash and merge to turn this in to a single commit (it's absolutely fine that you had multiple commits... a typical branch will see a number of commits to it).

PR #290, opened 2023-08-21
Task 1 - XR Files

danielhorton4001:
Sample file IM000001 can be opened and have a temporary rtss applied.
Files IM000002 through 4 cannot be opened.

Phil made the RTSTURCT function.

### DIFF
--- a/src/Model/MovingModel.py
+++ b/src/Model/MovingModel.py
@@ -41,11 +41,6 @@ def create_moving_model():
     if dataset[0].Modality == "CT":
         is_ct = True
 
-    # Determine if dataset is CR for grayscale inverting
-    is_cr = False
-    if dataset[0].Modality == "CR":
-        is_cr = True
-
     if 'WindowWidth' in dataset[0]:
         if isinstance(dataset[0].WindowWidth, pydicom.valuerep.DSfloat):
             window = int(dataset[0].WindowWidth)
@@ -92,7 +87,7 @@ def create_moving_model():
 
     if not moving_dict_container.has_attribute("scaled"):
         moving_dict_container.set("scaled", True)
-        pixel_values = convert_raw_data(dataset, False, is_ct, is_cr)
+        pixel_values = convert_raw_data(dataset, False, is_ct)
     else:
         pixel_values = convert_raw_data(dataset, True)
 
--- a/src/Model/MovingModel.py
+++ b/src/Model/MovingModel.py
@@ -41,6 +41,11 @@ def create_moving_model():
     if dataset[0].Modality == "CT":
         is_ct = True
 
+    # Determine if dataset is CR for grayscale inverting
+    is_cr = False
+    if dataset[0].Modality == "CR":
+        is_cr = True
+
     if 'WindowWidth' in dataset[0]:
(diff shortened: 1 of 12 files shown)


src/Model/MovingModel.py
```diff
@@ -41,6 +41,11 @@ def create_moving_model():
     if dataset[0].Modality == "CT":
         is_ct = True
 
+    # Determine if dataset is CR for grayscale inverting
```
sjswerdloff, line 44: see comment regarding Photometric Interpretation

src/Model/xrRtstruct.py
```diff
@@ -0,0 +1,165 @@
...
+from pydicom import dcmread
+from pydicom.dataset import FileMetaDataset, validate_file_meta
+from pydicom.uid import generate_uid, ImplicitVRLittleEndian
+from pydicom.tag import Tag
+from pathlib import Path
+from copy import deepcopy
+import datetime
+
+#AttributeError: 'Dataset' object has no attribute 'FrameOfReferenceUID'
+#removed 2 lines relating to this error, might need to be looked at if this is required
```
sjswerdloff, line 11: For the RT SS to be DICOM conformant (and can have an expectation of being usable by anything else), it will need to have *a* FrameOfReferenceUID.  Given that one is trying to tie the RT SS to the CR, one could generate a new UID for the FoR, insert that in to the RT SS, and update the CR object by adding the FoR UID to that.  Given that you know you will need to do this up front, it would be better to update the CR first (with a generated FoR UID), and update the SOP Instance UID of the CR (because it's a new object and not interchangeable with the original CR that lacks the FoR UID), and then use that FoR UID in the RT SS.  Having a DICOM object with a public element it wouldn't normally have tends to be reasonably interoperable.  Having a DICOM object without a public element that it is required to have is problematic.

src/Model/readxr.py
sjswerdloff, line 1: this seems like it would better to have in the test area, or if we want to have separate utilities.
It doesn't seem necessary for the application itself.

sjswerdloff (changes requested):
Please take a look at Photometric Interpretation.
It should allow you to simplify some of the code.

sjswerdloff:
Very good draft pull request.

sjswerdloff:
The pyradiomics issue is one for the maintainers to work out.
I've submitted a PR to pyradiomics to get them to fix the issue with metadata in pyradiomics 3.1.0.

sjswerdloff:
Please fetch and pull from upstream master (to address pyradiomics issue and the poetry.lock issue)

sjswerdloff:
Yeah!  All checks have passed, so the master branch is a known good starting point.
And... @danielhorton4001  your changes have all passed tests as well.
Please move forward on using Photometric Interpretation (and drop me a line if you would like any assistance).

src/Model/readxr.py
danielhorton4001: This file was supposed to be removed, I guess I forgot to do that. Sorry!

danielhorton4001:
Still need to fix up xrStruct.py, but everything else (including functionality with the other sample files) is in.

sjswerdloff:
> Still need to fix up xrStruct.py, but everything else (including functionality with the other sample files) is in.

please do this in a separate PR.

PR #292, opened 2023-08-31
Change DVH.CSV export: Change dvh output to percentvolume vs dose

AnnieAnNguyen:
The current DVH.CSV export has columns in 'centiGray' and rows in '%'.

Change output to columns in '%' (100%, 99.5%, 99.0%, .....) and rows in 'centiGray' (no decimals, normal rounding). This will produce a spreadsheet with a fixed number of columns irrespective of the prescribed radiation dose.

### DIFF
--- a/src/Model/CalculateDVHs.py
+++ b/src/Model/CalculateDVHs.py
@@ -187,7 +187,7 @@ def dvh2pandas(dict_dvh, patient_id):
                 else:
                     last_volume = trough_i
             if last_volume != -1:
-                dvh_roi_list.append(str(last_volume) + 'cGy')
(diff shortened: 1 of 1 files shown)


sjswerdloff (approved):
LGTM.
You might want to add some comments regarding the meaning of some of the algorithm and variables (peak, trough), etc. but I see that there is some interpolation going on, which was at the heart of our algorithm discussion.

sjswerdloff:
The new CSV files contain what I would refer to as Volume Dose Histograms, rather than Dose Volume Histograms (given that the axes have essentially been swapped).

PR #293, opened 2023-09-03
DICOM View Controller - Scrolling through CT layer + Drag and Drop CT image

AnnieAnNguyen:
1. Enable the scroll mouse / 2 finger scroll of the mouse tab scrolling through the CT layer (instead of up and down of a single CT picture)
2. Enable drag mode: Click the left mouse, then hold and drag to move the CT image

### DIFF
--- a/src/View/mainpage/DicomView.py
+++ b/src/View/mainpage/DicomView.py
@@ -8,8 +8,7 @@ from src.Controller.PathHandler import data_path
 class CustomGraphicsView(QtWidgets.QGraphicsView):
     def __init__(self, parent=None):
         super().__init__(parent)
-    
-    self.setDragMode(QtWidgets.QGraphicsView.ScrollHandDrag)
+        self.setDragMode(QtWidgets.QGraphicsView.ScrollHandDrag)
 
     def wheelEvent(self, event: QtGui.QWheelEvent):
         modifiers = event.modifiers()
--- a/src/View/mainpage/DicomView.py
+++ b/src/View/mainpage/DicomView.py
@@ -5,6 +5,20 @@ from src.Model.PatientDictContainer import PatientDictContainer
 from src.constants import INITIAL_ONE_VIEW_ZOOM
 from src.Controller.PathHandler import data_path
 
+class CustomGraphicsView(QtWidgets.QGraphicsView):
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+    def wheelEvent(self, event: QtGui.QWheelEvent):
+        modifiers = event.modifiers()
+        if modifiers == QtCore.Qt.ShiftModifier:
+            super().wheelEvent(event)
+        elif modifiers == QtCore.Qt.AltModifier:
+            super().wheelEvent(event)
+        elif modifiers == QtCore.Qt.ControlModifier:
+            event.ignore()
+        else:
+            event.ignore()
 
 class DicomView(QtWidgets.QWidget):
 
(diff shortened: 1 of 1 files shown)


src/View/mainpage/DicomView.py
```diff
@@ -53,6 +68,37 @@ def init_view(self):
...
         self.view.setBackgroundBrush(background_brush)
+    
+    def wheelEvent(self, event: QtGui.QWheelEvent):
+        modifiers = event.modifiers()
+        if modifiers == QtCore.Qt.ShiftModifier:
+            super().wheelEvent(event)
+        elif modifiers == QtCore.Qt.AltModifier:
+            super().wheelEvent(event)
+        elif modifiers == QtCore.Qt.ControlModifier:
+            angle = event.angleDelta() / 8
```
sjswerdloff, line 79: 8 and 15 are "magic" numbers.
Are these the recommended values in Qt example code?
Is there a URL reference that could be put in comments in this function that would point someone to the reason why eight and fifteen are appropriate values.
I found:
https://doc.qt.io/qt-6/qwheelevent.html#angleDelta
and it seems like the appropriate value might be different with a high resolution mouse.  A RadOnc might have a high resolution mouse...

This is probably OK for now, but it might need to be adjustable in some way in the future, e.g. by a configuration item by the user, or by somehow querying the system  to see if more information is available.

src/View/mainpage/DicomView.py
```diff
@@ -53,6 +68,37 @@ def init_view(self):
...
+            delta = angle.y() / 15
+            if delta > 0:
+                self.zoom_in()
+            else:
+                self.zoom_out()
+        else:
+            num_degrees = event.angleDelta() / 8
```
sjswerdloff, line 86: make magic numbers in to constants, and make the divisors here variables (with the initial value of the variable being the appropriate constant).  Use names that will help explain what they are for (or just put in the URL from my previous comment).

sjswerdloff (changes requested):
please use constants and variables instead of magic numbers (even though those are clearly and appropriately taken from the Qt documentation)

src/View/mainpage/DicomView.py
AnnieAnNguyen: Yes Stuart, I will modify the code

src/View/mainpage/DicomView.py
```diff
@@ -70,21 +70,24 @@ def init_view(self):
         self.view.setBackgroundBrush(background_brush)
     
     def wheelEvent(self, event: QtGui.QWheelEvent):
+        degree = 8
+        step = 15
+        #https://doc.qt.io/qt-6/qwheelevent.html#angleDelta
```
sjswerdloff, line 75: Great.  This gets the job done of clarifying where the values come from, and turning them in to variables.

src/View/mainpage/DicomView.py
```diff
@@ -70,21 +70,24 @@ def init_view(self):
...
+            angle = event.angleDelta() / degree
+            delta = angle.y() / step
             if delta > 0:
                 self.zoom_in()
             else:
                 self.zoom_out()
         else:
-            num_degrees = event.angleDelta() / 8
-            num_steps = num_degrees / 15
+            num_degrees = event.angleDelta() / degree
```
sjswerdloff, line 89: this suggests that the variable has "units" of angleDelta per degree (or alternatively, that the inverse has units of degrees per angleDelta).

src/View/mainpage/DicomView.py
```diff
@@ -70,21 +70,24 @@ def init_view(self):
...
+            num_steps = num_degrees / step
```
sjswerdloff, line 90: and this suggests that the variable has units of degrees per step (or alternatively, that it's inverse has units of steps per degree).
units are sometimes useful in the names of variables to help clarify their meaning.

sjswerdloff (approved):
This will be very helpful should we run in to high resolution mice in the future.

PR #295, opened 2023-10-11
Custom Windowing Slider

danielhorton4001:
Functionality of the slider incomplete.


src/View/mainpage/WindowingSlider.py
```diff
@@ -0,0 +1,297 @@
...
+        return index
+
+    def set_bars_from_window(self, window, level):
+        """
+        Triggered when the user selects a windowing preset.
+        Adjusts the values into the 0-2000 range, then updates the bars.
+        :param window: the window value of the preset
+        :param level: the level value of the preset
+        """
+        level = level + 1000
```
sjswerdloff, line 148: rather than a magic value, best to assign as a constant or a variable.  If this 1000 is specifically 1/2 of the "2000", then a variable that is 1/2 of that constant or variable would be appropriate

sjswerdloff (changes requested):
replace "magic numbers" with constants or variables (preferably with names that clarify their intent).
update tests.
Very good for a draft...

src/Model/Windowing.py
```diff
@@ -22,6 +23,21 @@ def windowing_model(text, init):
...
+def windowing_model_direct(level, window, init):
+    """
+        Function triggered when a window is selected from the menu,
+        or when the windowing slider bars are adjusted
+        :param level: The desired level
+        :param window: The desired window
+        :param init: list of bool to determine which views are chosen
+        """
+    patient_dict_container = PatientDictContainer()
+    moving_dict_container = MovingDictContainer()
```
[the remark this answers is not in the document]
danielhorton4001: patient_dict_container and moving_dict_container weren't used in windowing_model() after moving the code to windowing_model_direct(), but patient_dict_container still was still used in both.

src/View/mainpage/WindowingSlider.py
```diff
@@ -0,0 +1,297 @@
...
+        if index >= self.slider_density:
+            return self.slider_density-1
+        return index
+
+    def window_to_index(self, val):
+        """
+        Converts a window value to a slider index
+        :param val: a 0-2000 value
+        """
+        normalized_val = val / 2000
```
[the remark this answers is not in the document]
danielhorton4001: This was intended for windowing/level values, but after looking back that is both not correct for those values, and also not the correct place for them. I will replace them with the correct constant.

src/View/mainpage/WindowingSlider.py
```diff
@@ -251,9 +276,17 @@ def mouse_release(self, event):
...
-        level = 100
-        window = 100
-        # windowing_model_direct(level, window, send)
+
+        top_bar = self.index_to_window(self.top)
+        bottom_bar = self.index_to_window(self.bottom)
+
+        level = (top_bar + bottom_bar) * 0.5
+        window = 2 * (bottom_bar - level)
+        level = level - 1000
```
sjswerdloff, line 285: magic number... please use a constant or variable with meaningful name.
It's not clear 1000 is the value needed

src/View/mainpage/WindowingSlider.py
```diff
@@ -32,15 +32,20 @@ class WindowingSlider(QWidget):
     MAX_CLICK_DIST = 5
     # Minimum rendering height for bottom bar
     MIN_BOTTOM_INDEX = 4
+    # Window/Level consts
+    MAX_PIXEL_VALUE = 4096
+    LEVEL_OFFSET = 1000
```
sjswerdloff, line 37: why is there a level offset?  and why is the value 1000?  Perhaps this has to do with the values retrieved from Qt for the slider bars?

sjswerdloff (commented):
So far so good.

sjswerdloff (approved):

PR #297, opened 2023-10-12
Change DVH.CSV export: Change dvh output to percentvolume vs dose in BATCH Processing 

AnnieAnNguyen:
The current DVH.CSV export has columns in 'centiGray' and rows in '%'.

Change output to columns in '%' (100%, 99.5%, 99.0%, .....) and rows in 'centiGray' (no decimals, normal rounding). This will produce a spreadsheet with a fixed number of columns irrespective of the prescribed radiation dose.

Change the export in BATCH Processing


sjswerdloff (approved):
Are these new style DVH importing and displayed correctly?
If not, create a separate defect /issue.
This looks like the same code as non batch mode, which is appropriate but it suggests that a future refactoring to have the code in one place and called for each mode rather than having duplicated code (create separate issue.  Identify it as a refactoring opportunity

PR #298, opened 2023-10-20
Windowing histogram

danielhorton4001:
In 3D view, it syncs to the Axial view.
The seperate views cannot support their own windowing sliders until the way windowing is handled is changed to accomodate individual slice windowing.


sjswerdloff (approved):
I'd like a separate issue created to address the magic number.
I'll approve based on the assumption that will get fixed in a subsequent PR.

PR #299, opened 2023-10-21
Check index in range  before  incrementing

sjswerdloff:
apply guard, check bin index is within range  prior to indexing in to…histogram


didymo (approved):
Looks fine.

Issue #314, opened 2025-01-22
pymedphys and python version


didymo:
While attempting to update the Python and pysdie6 versions, I bumped into a small pymedphys issue.

poetry install --no-root
Updating dependencies
Resolving dependencies... (1.7s)

The current project's supported Python range (>3.9,<3.12) is not compatible with some of the required packages Python requirement:
  - pymedphys requires Python >=3.8,<3.11, so it will not be satisfied for Python >=3.11,<3.12

Because pymedphys[user] (0.39.3) requires Python >=3.8,<3.11
 and no versions of pymedphys match >0.39.3,<0.40.0, pymedphys is forbidden.
So, because onkodicom depends on pymedphys[user] (^0.39.3), version solving failed.

  * Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For pymedphys, a possible solution would be to set the `python` property to ">3.9,<3.11"

sjswerdloff:
I'll look at releasing a pymedphys 0.40, starting shortly after the 24th of January _cough_.
It has `python = ">=3.10,<3.13"`.

sjswerdloff:
> I'll look at releasing a pymedphys 0.40, starting shortly after the 24th of January _cough_. It has `python = ">=3.10,<3.13"`.

The changes made that will release in pymedphys 0.40 should also alleviate a number of chained dependency issues when something in OnkoDICOM attempts to pull in all dependencies regardless of whether they are actually being called by application code (pymedphys has a dynamic dependency approach for very late loading of projects so that not everything has to be in the environment).

didymo:
Nice

sjswerdloff:
pymedphys 0.40 was already released, what is about to come out is 0.41.
But even pointing to pymedphys 0.40 ran in to issues.
I'll need to dig a bit deeper to figure out what's going on (pymedphys ran in to problems with streamlit which has not maintained good backward compatibility...)

sjswerdloff:
PySide 6.3.0 is not compatible with python 3.11
pymedphys 0.40 is not compatible with python 3.9.
So that leaves python 3.10 for the PR action/tests.

sjswerdloff:
The test abort issues were due to changes in the QSpacerItem API (PySide 6, maybe 6.4)  
There are a number of API issues in matplotlib in Python 3.11 as well.
I think we should set a lower bound of 3.10, and then work to get 3.11 functioning (and then 3.12... given that 3.13 is in release).

[...]

PR #315, opened 2025-01-26
Fix deps pmp40 pydcm244

sjswerdloff:
Intended to address #314 
update reference to pymedphys to 0.40, and to install all extras, and specify a specific range of python versions
update reference to pyside6 to specify a range of python versions
update reference to pydicom to pin to version 2.4.4 (3.x has API changes that break dicompyler-core).
poetry.lock updated in unix environment:

```
uname -a
```

```
Darwin Stuarts-MacBook-Pro-2.local 23.6.0 Darwin Kernel Version 23.6.0: Fri Nov 15 15:13:15 PST 2024; root:xnu-10063.141.1.702.7~1/RELEASE_ARM64_T6000 arm64
```


didymo:
The conflicting poetry.lock file was deleted and rebuilt, then recommitted, and it still fails. Could you fix it on your end, please??

sjswerdloff:
> The conflicting poetry.lock file was deleted and rebuilt, then recommitted, and it still fails. Could you fix it on your end, please??

I wasn't seeing a complaint about the poetry.lock file, but I was seeing the core dump on the Ubuntu build (and *not* the Windows build, same minor version of python).  I don't have the problem on MacOS, so I'll clone in to Ubuntu on x86 (found my old Intel MacBook) and try to replicate.

sjswerdloff:
> > The conflicting poetry.lock file was deleted and rebuilt, then recommitted, and it still fails. Could you fix it on your end, please??
> 
> I wasn't seeing a complaint about the poetry.lock file, but I was seeing the core dump on the Ubuntu build (and _not_ the Windows build, same minor version of python). I don't have the problem on MacOS, so I'll clone in to Ubuntu on x86 (found my old Intel MacBook) and try to replicate.

Able to replicate on x86, the core dump happens while running test_pet_ct.py
Will continue to investigate

src/View/AddOnOptions.py
```diff
@@ -154,7 +154,10 @@ def set_layout(self):
         self.layout.addWidget(self.option_widget, 0, 1, 1, 3)
         try:
             # PySide6 worked on Windows
-            hspacer = QtWidgets.QSpacerItem(QtWidgets.QSizePolicy.Expanding,QtWidgets.QSizePolicy.Minimum)
+            hspacer = QtWidgets.QSpacerItem(1,
```
sjswerdloff, line 157: not clear what is an acceptable set of values for the first two arguments (preferred width and height, respectively).

src/View/addons/ImageFusionAddOnOption.py
```diff
@@ -67,13 +72,19 @@ def setupUi(self):
         self.gridLayout.setContentsMargins(0, 0, 0, 0)
         self.gridLayout.setVerticalSpacing(0)
 
+        hspacer = None
         # Create horizontal spacer in the middle of the grid
         try:
             # PySide6 worked on Windows
-            hspacer = QtWidgets.QSpacerItem(QtWidgets.QSizePolicy.Expanding,QtWidgets.QSizePolicy.Minimum)
+            hspacer = QtWidgets.QSpacerItem(5,
```
sjswerdloff, line 79: not clear what are appropriate preferred width and height values.

src/View/addons/ImageFusionAddOnOption.py
```diff
@@ -67,13 +72,19 @@ def setupUi(self):
...
+                                            16,
+                                            QtWidgets.QSizePolicy.Policy.Expanding,
+                                            QtWidgets.QSizePolicy.Policy.Minimum
+                                            )
         except TypeError:
             # PySide6 needed for Ubuntu 22.04 (perhaps due to PySide 6.4.0.1)
             # no idea if these are appropriate default values
+            logger.error("Type Error when defining spacer item")
             hspacer = QtWidgets.QSpacerItem(16,5)
```
sjswerdloff, line 88: these are magic numbers in the sense that there is no explanation of where they came from, nor are they explicit named constants.

test/test_view_dvh_tab.py
```diff
@@ -83,18 +86,54 @@ def test_object():
...
+
+
 def test_dvh_tab_with_dvh_not_calculated(qtbot, test_object, init_config):
     if test_object.main_window.dvh_tab.patient_dict_container.has_attribute("raw_dvh"):
         test_object.main_window.dvh_tab.patient_dict_container.additional_data.pop("raw_dvh")
     assert test_object.main_window.dvh_tab.patient_dict_container.has_attribute("raw_dvh") is False
     # check that Calculate DVH tab must appear
 
     test_object.main_window.show()
+    layout = test_object.main_window.dvh_tab.dvh_tab_layout.itemAt(1).layout()
```
sjswerdloff, line 106: The entire determination of the button_calc_dvh warrants a rewrite with an explanation of how it is being "discovered" and why the specific approach is being used.  Will this fail if there is a change in layout in the application?  Is there a more dynamic way of getting to the "right answer"?
This test was failing on MacOS 14.7 (Sonoma)/Apple Silicon, and it took these changes to get it to pass.

sjswerdloff (commented):
These comments are to aid in identifying the areas that I would like for more intensive review for this PR, and to suggest where additional work might be necessary.

didymo:
Thanks for your feedback. To date, I have focused on the conflicts within the poetry.lock file. It is difficult to describe the level of frustration with getting the lock file to resolve. I will look at your focus points over the next few days.

Thanks for the comments.

src/View/AddOnOptions.py
didymo: I am not in a position to comment at this time.

src/View/addons/ImageFusionAddOnOption.py
didymo: Agreed

src/View/addons/ImageFusionAddOnOption.py
didymo: Agreed

test/test_view_dvh_tab.py
didymo: Over time, with so many teams, there is much to rethink and refactor.

PR #322, opened 2025-02-10
Updates for py311

sjswerdloff:


pyproject.toml
```diff
@@ -47,6 +47,8 @@ scikit-learn = "*"
 imblearn = "*"
 kaplanmeier = "*"
 pyinstaller = "*"
+lifelines = "0.29.0"
```
sjswerdloff, line 50: lifelines is a dependency of kaplanmeier. the latest version of lifelines requires sciPy 2.x (which then triggers a lot of API failures).  Pinning at 0.29.0 moves it to the next to last version which is still compatible with everything else.  I don't recall if it required pandas 2.x, but that was probably the driver for updating the version of pandas (which then triggered the need to change the API call in Anonymise).

sjswerdloff:
Updating to SciPy 2.x is the next dependency that I will try to work on, although I could try python 3.12 to see if it's low hanging fruit.

pyproject.toml
didymo: agreed

PR #345, opened 2025-09-02
Add changes to master into dev branch

mattbrgn:
Merging the changes made to master branch into dev branch.


didymo:
Matt, What are your thoughts on Sourcery-ai comments?

Kahreiru:
> Matt, What are your thoughts on Sourcery-ai comments?

Mostly I copy pasted most the code from the copy pasted code in the existing examples in the codebase.
 - The platform detection method should probably be better than a string comparison but I'm not sure how else to do it.
 - There probably should be some form of error handling on the stylesheet. If it does have an issue it does raise an OSError and my assumption is that it wouldn't write anything to the memory if that happen then the class will raise a new error.
 - As for the single use function vs the Singleton comment. My option is that it would be better as it allows for better maintaince in the future if someone would like to add more functionality to the StyleSheetReader in the future. Also, it allows still maintaining the current API to the class where as a function would mean they would need to write an entirely new function would makes a new copy in memory or change the way the function works. Its also easier to find when it is not crammed into a util file.

PR #357, opened 2025-09-09
updated the viewer to allow for holes to be shown

AJB-BigA:
[Changed the way the polygons are added to the display scene to allow for the viewing of "holes" in drawing. This update only effects the viewing.

## Summary by Sourcery

Enhancements:
- Refactor ROI rendering to use QPainterPath with odd-even fill rule for displaying holes in polygons


AJB-BigA:
I removed the test that was giving me problem......... 
Not sure if this approach is appropriate but it only failed on Linux

PR #360, opened 2025-09-10
Changing more stylesheet reads to use StyleSheetReader

Kahreiru:
Improvements made:

Went though as much of the program as I could think of to to change style sheet reads to the StyleSheetReader class. which should reduce the number of open file streams which were never closed.
Added Logging to StyleSheetReader
When doing this I don't think I came across a single file stream that was opened using open() which was explicitly closed.


Kahreiru:
@sourcery-ai review

PR #373, opened 2025-09-19
Update to stylesheets, Made StyleSheetReader Singleton

Kahreiru:
Update to stylesheets, Made StyleSheetReader Singleton
Stylesheet update to work in dark mode setup - forced it to always be the same colours 

Made StyleSheetReader Singleton
So it always calls the same method call which is cached. As before it was singleton each get_stylesheet() call would have a different `self` input causing a new read of the style sheet.

## Summary by Sourcery

Improve stylesheet management by making StyleSheetReader a singleton, enhancing logging, and consistently applying the stylesheet across menu components

Enhancements:
- Convert StyleSheetReader to a singleton to cache stylesheet loading
- Add debug logging of the stylesheet file path when reading
- Apply the shared stylesheet to the main menu bar and its menu items including File, Tools, and Export

### DIFF
--- a/res/stylesheet-win-linux.qss
+++ b/res/stylesheet-win-linux.qss
@@ -1,11 +1,23 @@
 QMainWindow {
-    background-color: rgb(244, 245, 245);
+    background-color: #F5F4F5;
+}
+
+QWidget {
+    background-color: #F5F4F5;
+}
+
+QMenu {
+    color: #000000
+}
+
+QMenuBar {
+    color: #000000
 }
 
 QProgressBar {
     border-radius: 1px;
     background-color: #b299e6;
-    color: rgb(230, 230, 235);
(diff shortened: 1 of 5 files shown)


Kahreiru:
> - ActionHandler now uses StyleSheetReader but doesn’t import it, which will cause a NameError at runtime—please add the import.
This is imported in the file src/Controller/ActionHandler.py on line 9

> - You’re calling get_stylesheet() for each menu individually; consider fetching it once or setting it on a higher-level widget so 
It is a cached method call on a Singleton Object, so is it already set on a static level.

BurdManMan (approved):
LGTM. Can't wait to stop using light mode.

Kahreiru:
Just to be clear this isn't adding dark mode this is making it not change whether in dark mode or not

PR #384, opened 2025-09-24
Can we sync Dev to Master

Kahreiru:
I would like to be able to sync my branches without worrying if I'm syncing something which might break something


AJB-BigA:
no, I would like to push our draw ROI first then sync dev and master

Kahreiru:
if the commits exist on both it shouldn't matter and if we merge both in any order we are just going to get the same result

PR #411, opened 2025-10-23
Critical Bug Fix fixed image fusion on mac

Nowhere4Nothing:
Fixed the not loading with mac, used python threading
Fixed auto fusion not loading

## Summary by Sourcery

Add threaded loading and progress queue to Image Fusion UI to fix macOS loading issues and improve error handling

New Features:
- Introduce a progress queue polled by ImageFusionProgressWindow via QTimer
- Enable manual fusion loading in a background Python thread with interrupt flag and progress queue

Bug Fixes:
- Fix manual fusion loading hang on macOS by moving loader into a separate thread with proper progress reporting
- Fix auto fusion branch not loading by handling boolean result types and showing error dialogs
- Replace commonprefix with commonpath in MovingImageLoader to correctly compute shared directory

Enhancements:
- Unify progress_callback usage in ManualFusionLoader and MovingImageLoader as direct callback calls
- Add robust exception handling in MovingImageLoader.load with helper method to emit progress and errors
- Connect fusion mode radio buttons to revalidate and toggle the Confirm button upon mode changes

### DIFF
--- a/src/View/ImageFusion/ManualFusionLoader.py
+++ b/src/View/ImageFusion/ManualFusionLoader.py
@@ -78,11 +78,9 @@ class ManualFusionLoader(QtCore.QObject):
                 return
             self._load_with_vtk(main_thread_progress_callback)
         except Exception as e:
-            import traceback
-            stack = traceback.format_exc()
             main_thread_progress_callback(("Error loading images", e))
-            logging.exception("Error loading images: %s\n%s", e, stack)
-            self.signal_error.emit((False, f"{e}\n{stack}"))
+            logging.exception("Error loading images: %s\n%s", e)
+            self.signal_error.emit((False, f"{e}"))
 
     def _load_with_vtk(self, progress_callback):
         """
--- a/src/View/ImageFusion/ManualFusionLoader.py
+++ b/src/View/ImageFusion/ManualFusionLoader.py
@@ -1,7 +1,10 @@
+import time
+
 from PySide6 import QtCore, QtWidgets
 import logging
 import os
 import pydicom
+import platform
 import numpy as np
 from pydicom import dcmread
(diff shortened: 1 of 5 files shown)


src/View/ImageFusion/ManualFusionLoader.py
```diff
@@ -195,14 +204,15 @@ def _load_with_vtk(self, progress_callback):
...
                 self.signal_error.emit((False, f"Error extracting transform: {e}"))
                 return
 
-        # Do any overlay generation or heavy work here if needed
+        # Last message before overlay loaded
         if progress_callback is not None:
-            progress_callback.emit(("Preparing overlays...", 90))
+            progress_callback(("Preparing overlays...", 90))
             QtCore.QCoreApplication.processEvents()
+            time.sleep(0.05)
```
Kahreiru: while it is only 0.05 seconds. Is this a thread blocking? could a non-thread blocking option be used here

Kahreiru (approved):
look ok provided things which are mentioned but not a big deal

src/View/ImageFusion/ManualFusionLoader.py
Nowhere4Nothing: it doesn't need to be there at all it just goes too quick and would go 50% to done, was just thinking it would be better for the user to see it go to 90% for a very quick time before loading. can be removed if best

sjswerdloff (approved):
I created an issue for the renaming of the helper method.
While the bot might consider this to be a nit, this type of thing prevents abstraction.
In other words, I can't read the code calling it and have any idea of what's going to happen, so I have to go *find* the function and read it.  So it doesn't really help me as someone who needs to understand the code, you might as well have just copy/pasted it where-ever it was being used.

